PR #36, opened 2017-01-30
src,build: improve the native module subsystem

aqrln:
* [x] Split `jsrs-impl.cc` into separate modules and make stylistic fixes.
* [x] Rename the native addon to `jstp` since there already is a function that is not a part of JSRS.
* [x] Fix `binding.gyp`: make `cflags` not ignored on macOS (as it appeared they used to be) and do not use `-O3` in Debug configuration.
* [x] Make `v8::String` instantiations and throwing V8 exceptions macros to avoid boilerplate code.


src/jsrs_serializer.cc
```diff
@@ -0,0 +1,211 @@
...
+  return key;
+}
+
+// Checks if a string can be a non-quoted object key.
+static bool IsValidKey(Isolate* isolate, const String::Utf8Value& key) {
+  bool result = true;
+  const char* key_str = *key;
+  for (int i = 0; i < key.length(); i++) {
+    if (key_str[i] == '_') continue;
+    if ((i == 0 && !isalpha(key_str[i])) || !isalnum(key_str[i])) {
```
belochub: Functions `std::isalpha` and `std::isalnum` are both defined in `<cctype>` header which is not included in this file.

src/packet_parser.cc
```diff
@@ -0,0 +1,64 @@
...
+                                  const String::Utf8Value& in,
+                                  Local<Array> out) {
+  size_t total_size = 0;
+  size_t parsed_size = 0;
+  const char* source = PrepareString(isolate, *in, in.length(), &total_size);
+  const char* curr_chunk = source;
+  int index = 0;
+
+  while (parsed_size < total_size) {
+    auto chunk_size = strlen(curr_chunk);
```
belochub: `std::strlen` is defined in `<cstring>` which is not included.

src/node_bindings.cc
```diff
@@ -0,0 +1,88 @@
...
+  }
+  if (!args[0]->IsString() || !args[1]->IsArray()) {
+    THROW_EXCEPTION(TypeError, "Wrong argument type");
+    return;
+  }
+
+  HandleScope scope(isolate);
+
+  String::Utf8Value str(args[0]->ToString());
+  auto array = Local<Array>::Cast(args[1]);
```
belochub: I guess it is better to use `v8::Local::As` here to keep cast style consistent, but this one is optional and completely up to you.

src/jsrs_serializer.cc
```diff
@@ -0,0 +1,211 @@
...
+      static_cast<int>(result_str.size())).ToLocalChecked();
+}
+
+// Returns a string representing an escaped control character.
+// If the given character is not a control one, returns nullptr.
+// The `size` is being incremented by the length of the resulting
+// string, but always at least by two, even when the function
+// returns nullptr.
+static const char* GetEscapedControlChar(char str, size_t* size) {
+  static const char* control_chars[0x20] = {
```
aqrln: @belochub and get a `warning: ISO C++11 does not allow conversion from string literal to
      'char *const'` 😉

src/jsrs_serializer.cc
aqrln: Fixed in 07eafcc7749b6af8818f8771e788a4571846c0b5

src/packet_parser.cc
aqrln: Ditto

src/jsrs_serializer.cc
aqrln: 330caf9dcdc47bbfbfc14cb63834b5b671438457

src/node_bindings.cc
aqrln: @belochub a103c2e0a974ffb85eeb98541abdcea993d0bc3e

aqrln:
@belochub I've added new commits that fix mentioned nits and issues. Does this LGTY?

aqrln:
@belochub I don't think we need separate macros to create strings. It was only `isolate->ThrowException` together with string instantiation thing that created lots of boilerplate code, and that its handled by a single macro, so I mark the last item as completed. What do you think?

belochub:
@aqrln, yes, I think at can be marked as completed. Changes LGTM.

aqrln:
Landed in 56c57fe588957176f763d7be07e8e29abcfc7e3d.

aqrln:
@belochub do you know why was this marker as minor and not patch?

belochub:
@aqrln, I don't remember the actual reason.

aqrln:
Ah, doesn't matter for now. Minor and patch are the same till 1.0. Marked the backport PR as minor though too.

PR #37, opened 2017-01-31
build: compile in ISO C++11 mode

aqrln:
By default `node-gyp` uses `-std=gnu++0x`. This commit modifies `cflags` to use `-std=c++11`.

@belochub there's a question we need to resolve. For the sake of simplicity I placed `-std=c++11` to `cflags` GYP variable that is applied both to C and C++ sources. Currently it has no side-effects since we have no pure-C sources, and most likely won't have such. Technically it would be more correct to place it into `cflags_cc` and make `cflags_cc` inherit `cflags` explicitly (because strangely `cflags` change their behavior and get ignored in generated C++ flags in `*.target.mk` if there are `cflags_cc` present in `binding.gyp`), but it will result in quite more complex `binding.gyp`.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,8 +1,13 @@
 {
   'variables': {
-    'jstp_base_cflags': ['-Wall', '-Wextra', '-Wno-unused-parameter'],
-    'jstp_debug_cflags': ['-g', '-O0'],
-    'jstp_release_cflags': ['-O3']
+    'jstp_base_ccflags': [
+      '-Wall',
+      '-Wextra',
+      '-Wno-unused-parameter',
+      '-std=c++11'
+    ],
+    'jstp_debug_ccflags': ['-g', '-O0'],
+    'jstp_release_ccflags': ['-O3']
   },
   'targets': [
     {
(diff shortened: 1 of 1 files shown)


belochub:
@aqrln, maybe it is possible to completely avoid providing `cflags` and provide `cflags_cc` instead, if it is not likely for us to have pure-C sources in the future?

aqrln:
@belochub yeah, that sounds reasonable.

aqrln:
@belochub updated

belochub:
@aqrln LGTM

aqrln:
CI passing (which is good since I only tested on Clang, not GCC), landed in 08516fd290f60e416123a9f6dd6ea7b4622a3bca.

aqrln:
@belochub btw, are there versions of GCC (and if there are, which is the last of them) that support `-std=gnu++0x` but do not support `-std=c++11`?

belochub:
@aqrln, yes, there are: GCC from 4.3 till 4.6 only had option `-std=c++0x` (or `-std=gnu++0x`), but not `-std=c++11`.
Also, according to GCC official site:
> GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard

Which means that it is better to avoid compiling C++11 programs on GCC versions earlier than 4.8.1 because even with `-std=c++11` option enabled it does not guarantee that all of the needed features will work.

aqrln:
@belochub, ah, and GCC 4.8+ is a requirement for building Node.js addons. So this change will not break a build on some Debian or CentOS 6, which I was afraid of.

PR #40, opened 2017-01-31
build: improve error handling

aqrln:
* Handle the situation when a compiler cannot be spawned and `exit` event is not fired.
* Make `0` a named constant.


belochub:
LGTM

PR #41, opened 2017-01-31
lib: refactor record-serialization.js

aqrln:
Get rid of nested try-catch statements. Decouple `safeRequire` function that can be used later.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -25,3 +25,14 @@ common.forwardMultipleEvents = (from, to, eventNames) => {
     common.forwardEvent(from, to, event);
   });
 };
+
+// Try to require `moduleName` and return the exported object if the module is
+// found or null otherwise.
+//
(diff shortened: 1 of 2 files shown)


lib/common.js
```diff
@@ -25,3 +25,14 @@ common.forwardMultipleEvents = (from, to, eventNames) => {
     common.forwardEvent(from, to, event);
   });
 };
+
+// Try to require `moduleName` and return the exported object if the module if
```
belochub: There is a typo here, probably should be `...if the module is found...`

aqrln: @belochub thanks!

aqrln:
@belochub I updated the commit. PTAL.

aqrln:
Landed in 2c5e74eaa8de635819d040269867f8dbe8450aa1.

PR #44, opened 2017-02-01
parser: fix possible memory leaks

aqrln:
Fix possible memory leaks in `ParseString` function from `jsrs_parser.cc`.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -372,6 +372,7 @@ Local<Value> ParseString(Isolate*    isolate,
   }
 
   if (!is_ended) {
+    delete[] result;
     THROW_EXCEPTION(SyntaxError, "Error while parsing string");
     return String::Empty(isolate);
   }


src/jsrs_parser.cc
```diff
@@ -360,6 +360,8 @@ Local<Value> ParseString(Isolate*    isolate,
       char* symb =
           GetControlChar(isolate, begin + ++i, &out_offset, &in_offset);
       if (!symb) {
+        delete[] result;
```
belochub: There is no point in calling `delete[]` on `nullptr`, and that is what you are doing here.
Also, about TODO: no need to throw an exception, because `GetControlChar` already takes care of this and throws when it returns `nullptr`.

aqrln: @belochub uh, didn't look into `GetControlChar`.

aqrln:
@belochub updated the commit. Can I get a sign-off?

aqrln:
Landed in 534eb219314877201b72d420807a72ab194d8756.

PR #52, opened 2017-02-06
connection: handle optional callbacks properly

aqrln:
For some public methods of Connection that perform network actions
callbacks are optional. This commit fixes the behavior of default
actions when a callback function is not passed so that (1) the system
logic is always run and (2) a Connection doesn't complain about
callback packets for unknown source packets.


aqrln:
Landed in 170e6d685d50f30977120205d68bd89da24ca45b.

Issue #35, opened 2017-01-18
Reconnecting to existing sessions


aqrln:
Now we have an abstraction of session but it isn't used anywhere. There must be a possibility to connect to an existing session specifying its ID in handshake packet. When a client loses its connection, it must reconnect to the session automatically and transparently.

aqrln:
In order to implement this, we should change the format of `handshake` packets so it is not strictly bound to login and password but rather allows to specify the authentication strategy (login/password, anonymous, existing session etc). I am working on an implementation currently.

belochub:
Closed in https://github.com/metarhia/jstp/commit/4d600f4587238f0a8165bbb812e4570c5a9f01c7.

PR #54, opened 2017-02-06
protocol: change the format of handshake packets

aqrln:
* Introduce the concept of authentication strategies.
* Implement `login` and `anonymous` strategies.
* Make handshake packets include the strategy as an action field.
* Update tests.
* Update docs.

Refs: https://github.com/metarhia/JSTP/issues/35#issuecomment-277692315

See more details about the implementation in the updated `doc/protocol.md`.

### DIFF
--- a/doc/protocol.md
+++ b/doc/protocol.md
@@ -129,10 +129,22 @@ Handshake packets always have ID equal to `0`. The response contains either
 the key `ok` with a value that is the session identifier or `error` that is
 an array with error code and optional error message.
 
+The action field of handshake requests specifies the authentication strategy.
+There are two supported strategies now:
+
+* `login` — authentication with login and password. The payload is an array of
+  two elements: username and password, both represented as strings.
+* `anonymous` — anonymous session request. The payload is ignored (e.g., `true`
+  or an empty array may be used). For anonymous handshakes the action field can
+  be omitted completely, `anonymous` is implied by default.
+
+More strategies may be added in the future (for example, `session` to reconnect
+to an existing session after connection break due to a network error).
+
 Successful handshake:
 
 ```javascript
-C: {handshake:[0,'example'],marcus:'7b458e1a9dda....67cb7a3e'}
+C: {handshake:[0,'example'],login:['marcus','7b458e1a9dda....67cb7a3e']}
 S: {handshake:[0],ok:'9b71d224bd62...bcdec043'}
 ```
 
@@ -153,7 +165,7 @@ Successful handshake of [Impress](https://github.com/metarhia/Impress) worker
 connecting to a private cloud controller:
 
 ```javascript
-C: {handshake:[0,'impress'],S1N5:'d3ea3d73319b...5c2e5c3a'}
+C: {handshake:[0,'impress'],login:['S1N5','d3ea3d73319b...5c2e5c3a']}
 S: {handshake:[0],ok:'PrivateCloud'}
 ```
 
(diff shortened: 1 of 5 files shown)


lib/connection.js
```diff
@@ -388,26 +390,26 @@ Connection.prototype._processHandshakeRequest = function(packet, keys) {
...
   });
 
-  const callback = this._onSessionCreated.bind(this, username);
-  this.server.startSession(this, application, username, password, callback);
+  this.server.startSession(this, application, authStrategy, credentials,
+    this._onSessionCreated.bind(this));
 };
 
 // Callback of authentication operation
 //   username - user login, if any
```
belochub: You can probably change order of arguments here as well, to correspond with function signature.

lib/simple-auth.js
```diff
@@ -14,20 +14,23 @@ module.exports = simpleAuth;
 // Start session. Only anonymous handshakes are allowed.
 //   connection - JSTP connection
 //   application - application instance
-//   username - null (required by the auth callback contract)
-//   password - null (required by the auth callback contract)
+//   strategy - authentication strategy (only 'anonymous' is supported)
+//   credentials - authentication credentials
 //   callback - callback function that has signature (error, sessionId)
```
belochub: Callback signature here does not correspond to the callback you are calling on line 33.

lib/simple-auth.js
```diff
@@ -14,20 +14,23 @@ module.exports = simpleAuth;
...
 //
-simpleAuth.startSession = (connection, application,
-                           username, password, callback) => {
-  if (username) {
+simpleAuth.startSession = (
+  connection, application,
+  strategy, credentials, callback
+) => {
+  console.dir({ strategy, credentials });
```
belochub: Was it intentionally left here?

aqrln: Leftover of fixing tests. Thanks for noticing it.

doc/protocol.md
```diff
@@ -129,10 +129,22 @@ Handshake packets always have ID equal to `0`. The response contains either
 the key `ok` with a value that is the session identifier or `error` that is
 an array with error code and optional error message.
 
+The action field of handshake requests specifies the authentication strategy.
+There are two supported strategies now:
+
+* `login` — authentication with login and password. The payload is an array of
+  two elements: user name and password, both represented as strings.
```
belochub: In this case it is usually more appropriate to write "username", not "user name".

aqrln:
@belochub updated to resolve the issues and nits.

aqrln:
Rebased to resolve conflicts.

aqrln:
@tshemsedinov awaiting your review.

aqrln:
@tshemsedinov it's semver-major anyway, so it will go along with any other breaking changes. All non-breaking changes are still being backported to v0.6.

aqrln:
@tshemsedinov and it is still possible to do a protocol-level compatibility fixup for v0.6 granted we have no users named 'login', 'anonymous' or 'session' :)
Whom we actually don't even if someone used such nicknames since the protocol-level authentication is not used in the current projects and all users are anonymous from the point of view of JSTP.

aqrln:
@tshemsedinov @belochub it has just occurred to me that in fact the protocol-level compatibility will not be broken for our applications because of the reasons I told in the previous comment. Since only anonymous handshakes were used and authentication was perform using RPC methods, and anonymous handshakes syntax has not changed, we can update the server to JSTP 0.7 without updating the clients.

aqrln:
@tshemsedinov @belochub so basically we don't even need backporting this PR to v0.6 and supporting v0.6 at all.

aqrln:
If we introduce any new breaking changes that will not be feasible to work around, we'll release v0.7 in its current state, call new breaking changes v0.8 and continue landing non-breaking changes to v0.7.

aqrln:
Landed in 9e55d3c11d3be63e49b97485065c9737957f708d.

PR #59, opened 2017-02-07
server: clean internal structures on close

aqrln:
After server has been closed, clean internal structures such as `Server#clients` and `Server#_cachedClientsArray`.


aqrln:
Landed in 770d4522ac2bf81c67a8e531de8759a59c4055c9.

Issue #60, opened 2017-02-08
Make parser single-pass


aqrln:
Avoid extra memory copying by eliminating `jstp::parser::internal::PrepareString` and extracting lexems on demand instead.

aqrln:
Avoid extra memory copying by eliminating `jstp::parser::internal::PrepareString` and extracting lexems on demand instead.

PR #61, opened 2017-02-08
parser: make parser single-pass

belochub:
* Eliminate `jstp::parser::internal::PrepareString` function thus avoiding extra memory copying.
* Fix deleting comment blocks located inside of strings.
* Fix key parsing not throwing error when it included spaces but was not enclosed in quotes.

Fixes: #60

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -7,12 +7,14 @@
 #include <cstddef>
 #include <cstdlib>
 #include <cstring>
+#include <functional>
 #include <vector>
 
 #include "common.h"
 #include "unicode_utils.h"
 
 using std::atof;
+using std::function;
 using std::isalnum;
 using std::isalpha;
 using std::isdigit;
@@ -74,28 +76,30 @@ static constexpr Local<Value> (*kParseFunctions[])(Isolate*,
 };
 
 Local<Value> Parse(Isolate* isolate, const String::Utf8Value& in) {
-  size_t size;
-  const char* to_parse =
-      internal::PrepareString(isolate, *in, in.length(), &size);
-  if (!to_parse) {
-    return Undefined(isolate);
-  }
+  const char* str = *in;
+  const size_t length = in.length();
+  const char* end = str + length;
 
   Type type;
-  if (!GetType(to_parse, to_parse + size, &type)) {
+
+  size_t start_pos = internal::SkipToNextToken(str, end);
+  if (!GetType(str + start_pos, end, &type)) {
     THROW_EXCEPTION(TypeError, "Invalid type");
     return Undefined(isolate);
   }
 
   size_t parsed_size = 0;
   Local<Value> result =
-      (kParseFunctions[type])(isolate, to_parse, to_parse + size, &parsed_size);
-  if (size != parsed_size) {
+      (kParseFunctions[type])(isolate, str + start_pos, end, &parsed_size);
+
+  parsed_size += internal::SkipToNextToken(str + start_pos + parsed_size, end);
+  parsed_size += start_pos;
+
+  if (length != parsed_size) {
     THROW_EXCEPTION(SyntaxError, "Invalid format");
     return Undefined(isolate);
   }
 
-  delete[] to_parse;
   return result;
 }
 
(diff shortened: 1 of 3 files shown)


src/jsrs_parser.cc
```diff
@@ -74,28 +76,30 @@ static constexpr Local<Value> (*kParseFunctions[])(Isolate*,
...
-    return Undefined(isolate);
-  }
+  const char* str = *in;
+  const size_t length = in.length();
+  const char* end = str + length;
 
   Type type;
-  if (!GetType(to_parse, to_parse + size, &type)) {
+
+  size_t start_pos = internal::SkipToNextToken(str, end);
```
aqrln: To be technically correct, it should probably be `ptrdiff_t`, not `size_t`. Though they are the same things on x86/x86_64 and most other platforms.

src/jsrs_parser.cc
```diff
@@ -74,28 +76,30 @@ static constexpr Local<Value> (*kParseFunctions[])(Isolate*,
...
     return Undefined(isolate);
   }
 
   size_t parsed_size = 0;
   Local<Value> result =
-      (kParseFunctions[type])(isolate, to_parse, to_parse + size, &parsed_size);
-  if (size != parsed_size) {
+      (kParseFunctions[type])(isolate, str + start_pos, end, &parsed_size);
+
+  parsed_size += internal::SkipToNextToken(str + start_pos + parsed_size, end);
```
aqrln: Why not increment `str` pointer after each `SkipToNextToken` call? We don't need the initial address anyway.

src/jsrs_parser.cc
```diff
@@ -144,65 +148,76 @@ static bool GetType(const char* begin, const char* end, Type* type) {
...
+  }
+  return false;
+}
+
+// Returns count of bytes needed to skip to current comment ending.
+size_t SkipToCommentEnd(const char* str, const char* end) {
+  bool is_single_line;
+
+  switch (str[1]) {
+    case '/': {
```
aqrln: No need for braces.

src/jsrs_parser.cc
```diff
@@ -144,65 +148,76 @@ static bool GetType(const char* begin, const char* end, Type* type) {
...
+      is_single_line = true;
+      break;
     }
+    case '*': {
```
aqrln: Ditto

src/jsrs_parser.cc
```diff
@@ -144,65 +148,76 @@ static bool GetType(const char* begin, const char* end, Type* type) {
...
+      is_single_line = false;
+      break;
+    }
+    default: { // In case it is not a comment start
```
aqrln: 1. No need for braces.
2. Two spaces before a single-line comment.

src/jsrs_parser.cc
```diff
@@ -144,65 +148,76 @@ static bool GetType(const char* begin, const char* end, Type* type) {
...
 
-      if (comment_mode == kDisabled) {
-        if (IsWhiteSpaceCharacter(str + i, &size) ||
-            IsLineTerminatorSequence(str + i, &size)) {
-          i += size - 1;
-        } else {
-          result[j++] = str[i];
-        }
-      }
+  if (is_single_line) {
```
aqrln: Consider using a ternary operator, though it's up to you.

src/jsrs_parser.cc
belochub: I just followed Google C++ Styleguide.

src/jsrs_parser.cc
belochub: Ditto.

src/jsrs_parser.cc
aqrln: Well, the Google C++ Styleguide states:

> case blocks in switch statements can have curly braces or not, depending on your preference.

Then we need to decide whether we use them or don't.

belochub: I liked it more with braces, what is your opinion on this topic?

src/jsrs_parser.cc
belochub: For two reasons: it is `const char*` and to enable universal check on line 98.

src/jsrs_parser.cc
aqrln: IDK, I would not use them and have never used them before in such cases, but I checked it out in my editor and this code snippet indeed seems to look prettier and more readable with braces.

belochub: Let's stick with braces then. We need to update it in other switch-statements, should I apply it in this pull request, or it is the case for another one?

aqrln: Better raise a new PR.

src/jsrs_parser.cc
belochub: Here it is not the case, because `SkipToNextToken` returns `size_t`, which is fine (there is no pointer subtraction to get result in this function), but generally you are right and there are other occurrences where it can be important. I guess, I will still review and check for any issues that may be caused by this.

aqrln:
@belochub I don't feel well now so I will review the rest of the PR tomorrow.

src/jsrs_parser.cc
belochub: I investigated this topic and found out that the main difference between `size_t` and `ptrdiff_t` is that former is unsigned integer, whereas latter is signed.
Also, there are a lot of arguments on this topic and generally it is recommended to use signed types for array indexing as well as pointer arithmetic.
At the same time Google C++ Style Guide says:

> When appropriate, you are welcome to use standard types like size_t and ptrdiff_t.
> 
> We use int very often, for integers we know are not going to be too big, e.g., loop counters. Use plain old int for such things. You should assume that an int is at least 32 bits, but don't assume that it has more than 32 bits. If you need a 64-bit integer type, use int64_t or uint64_t.

I guess it will be better to substitute almost all `size_t` uses with `ptrdiff_t` or plain `int`, but we need to discuss which one will suit us better.
Also I think we need new issue and corresponding PR for this topic, so that we can talk it over and weigh all the pros and cons.

belochub:
Rebased on master and fixed indentation in single-line comment.

aqrln:
@belochub mind changing `Refs` to `Fixes` in the commit metadata?

src/jsrs_parser.cc
```diff
@@ -367,15 +380,23 @@ Local<Value> ParseString(Isolate*    isolate,
...
+      if (IsLineTerminatorSequence(begin + i + 1, &in_offset)) {
+        i += in_offset;
+      } else {
+        char* symb =
+            GetControlChar(isolate, begin + ++i, &out_offset, &in_offset);
+        if (!symb) {
+          return String::Empty(isolate);
+        }
+        strncpy(result + res_index, symb, out_offset);
+        delete[] symb;
```
aqrln: We can allocate a buffer on stack and pass it to `GetControlChar` by pointer.

src/jsrs_parser.cc
```diff
@@ -524,25 +545,21 @@ Local<String> ParseKeyInObject(Isolate*    isolate,
   } else {
     size_t current_length = 0;
     for (size_t i = 0; i < *size; i++) {
-      if (begin[i] == ':') {
+      if (begin[i] == '_' || (i != 0 ?
```
aqrln: Maybe decouple the check into an inline function?

src/jsrs_parser.cc
```diff
@@ -524,25 +545,21 @@ Local<String> ParseKeyInObject(Isolate*    isolate,
...
         if (current_length != 0) {
           result = String::NewFromUtf8(isolate, begin,
                                        NewStringType::kInternalized,
                                        static_cast<int>(current_length))
                                            .ToLocalChecked();
           break;
         } else {
-          THROW_EXCEPTION(SyntaxError, "Unexpected token :");
+          THROW_EXCEPTION(SyntaxError, "Unexpected identifier");
           return Local<String>();
```
aqrln: I'd like it to be consistent. In some cases we use `Local<String>`, in some cases — `String::Empty`. Though it is the matter of a different PR since it was not introduced in this one.

src/jsrs_parser.cc
```diff
@@ -579,20 +596,23 @@ Local<Value> ParseObject(Isolate*    isolate,
...
         break;
       }
       current_key = ParseKeyInObject(isolate,
                                      begin + i,
                                      end,
                                      &current_length);
       i += current_length;
+      i += SkipToNextToken(begin + i, end);
+      if (begin[i] != ':') {
+        THROW_EXCEPTION(SyntaxError, "Unexpected token");
```
aqrln: Maybe change the message to the more meaningful `"Colon expected"`?

src/packet_parser.cc
```diff
@@ -29,9 +29,10 @@ namespace packet_parser {
 Local<String> ParseNetworkPackets(Isolate* isolate,
                                   const String::Utf8Value& in,
                                   Local<Array> out) {
-  size_t total_size = 0;
+  size_t total_size = in.length();
   size_t parsed_size = 0;
-  const char* source = PrepareString(isolate, *in, in.length(), &total_size);
+  // const char* source = PrepareString(isolate, *in, in.length(), &total_size);
```
aqrln: Commented-out line should be deleted.

src/packet_parser.cc
```diff
@@ -29,9 +29,10 @@ namespace packet_parser {
 Local<String> ParseNetworkPackets(Isolate* isolate,
                                   const String::Utf8Value& in,
                                   Local<Array> out) {
-  size_t total_size = 0;
+  size_t total_size = in.length();
```
aqrln: I think I can now be `const size_t`.

src/jsrs_parser.cc
aqrln: Not related to this PR, though.

aqrln:
LGTM if the commented line is deleted. Other points are discussable.

belochub:
@aqrln updated to resolve some of the issues you addressed.

aqrln:
@belochub other issues mentioned have to be extracted and posted on the issue tracker. Will you do it or I should take care of it?

belochub:
@aqrln, I already have an idea for a new PR fixing some of the issues, so I guess I will also extract other issues as well.

belochub:
Landed in 3b2c1d311dbf50a669366688642929a3682940cc.

PR #62, opened 2017-02-09
src: add curly braces in `switch` statements

belochub:


belochub:
Landed in 49ac6a3f85b56bb7ef33cbe74646d4f0405c910d.

PR #66, opened 2017-02-10
parser: improve string parsing

belochub:
* Eliminate unnecessary assignments.
* Encode control characters in place to avoid copying and memory allocations.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -350,10 +350,11 @@ Local<Value> ParseIntegerNumber(Isolate*    isolate,
   return Integer::New(isolate, value);
 }
 
-static char* GetControlChar(Isolate*    isolate,
-                            const char* str,
-                            size_t*     res_len,
-                            size_t*     size);
+static bool GetControlChar(Isolate*    isolate,
+                           const char* str,
+                           size_t*     res_len,
+                           size_t*     size,
+                           char*       write_to);
 
 Local<Value> ParseString(Isolate*    isolate,
                          const char* begin,
@@ -361,7 +362,6 @@ Local<Value> ParseString(Isolate*    isolate,
                          size_t*     size) {
   *size = end - begin;
   char* result = new char[*size + 1];
-  memset(result, 0, *size + 1);
 
   enum { kApostrophe = 0, kQMarks} string_mode = (*begin == '\'') ?
                                                  kApostrophe :
@@ -375,7 +375,6 @@ Local<Value> ParseString(Isolate*    isolate,
         (string_mode == kApostrophe && begin[i] == '\'')) {
       is_ended = true;
       *size = i + 1;
-      result[res_index] = '\0';
       break;
     }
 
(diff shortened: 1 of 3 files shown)


src/jsrs_parser.cc
```diff
@@ -350,18 +350,18 @@ Local<Value> ParseIntegerNumber(Isolate*    isolate,
...
 
-static char* GetControlChar(Isolate*    isolate,
-                            const char* str,
-                            size_t*     res_len,
-                            size_t*     size);
+static bool GetControlChar(Isolate*    isolate,
+                           const char* str,
+                           size_t*     res_len,
+                           size_t*     size,
+                           char* write_to);
```
aqrln: Please align the parameter to match the surrounding style.

src/jsrs_parser.cc
```diff
@@ -375,21 +375,18 @@ Local<Value> ParseString(Isolate*    isolate,
...
     }
 
     if (begin[i] == '\\') {
       if (IsLineTerminatorSequence(begin + i + 1, &in_offset)) {
         i += in_offset;
       } else {
-        char* symb =
-            GetControlChar(isolate, begin + ++i, &out_offset, &in_offset);
-        if (!symb) {
+        bool ok = GetControlChar(isolate, begin + ++i, &out_offset, &in_offset,
```
aqrln: So that doesn't need copying at all, even to a stack-allocated buffer. Good catch!

src/jsrs_parser.cc
```diff
@@ -375,21 +375,18 @@ Local<Value> ParseString(Isolate*    isolate,
...
+                                 result + res_index);
+        if (!ok) {
```
aqrln: What do you think about making it one-line?

src/jsrs_parser.cc
```diff
@@ -418,51 +415,51 @@ static unsigned int ReadHexNumber(const char* str, size_t len, bool* ok);
...
-                            size_t*     res_len,
-                            size_t*     size) {
-  char* result = new char[5];
+// character and writes it to `write_to`.
+// Returns true if no error occured, false otherwise.
+static bool GetControlChar(Isolate*    isolate,
+                           const char* str,
+                           size_t*     res_len,
+                           size_t*     size,
+                           char* write_to) {
```
aqrln: The same alignment issue as above.

src/jsrs_parser.cc
```diff
@@ -418,51 +416,51 @@ static unsigned int ReadHexNumber(const char* str, size_t len, bool* ok);
...
-      *result = '\t';
+      *write_to = '\t';
       break;
     }
     case 'v': {
-      *result = '\v';
+      *write_to = '\v';
       break;
     }
     case '0': {
```
aqrln: *(Not related to this PR)*
Do we really need `'\0'` as a special case here? Either we support any octal sequences, or we don't support them at all.

belochub:
Added fix for possible memory leak in this string parsing.

src/jsrs_parser.cc
belochub: I think, that way it is more understandable than just
```c++
if (!GetControlChar(...)) {
```
Compiler should optimize it anyway.

aqrln: Never mind, for some reason it seemed to me that there's one short statement inside the `if` body, so I proposed to eliminate braces.

belochub:
@aqrln updated with alignment fixes.

src/jsrs_parser.cc
belochub: Yeah, I guess we don't.

belochub: Will make a new PR for this.

belochub:
Landed in 779965d4e58d0adf4b20c0c780567e7bdbdfdf32.

PR #68, opened 2017-02-10
parser: remove special case for '\0' literal

belochub:
Remove special case for '\0' in control character parsing.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -452,10 +452,6 @@ static char* GetControlChar(Isolate*    isolate,
       *result = '\v';
       break;
     }
-    case '0': {
-      *result = '\0';
-      break;
-    }
 
     case 'x': {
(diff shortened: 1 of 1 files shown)


aqrln:
Oh, sweet. Looks like we don't have tests for escape sequences.

aqrln:
Well, I think we can be pretty sure it doesn't introduce any regression since the change is trivial. But we must put writing proper tests for existing code on our agenda and do it ASAP. Since that, each PR will be required to add tests for any new or updated functionality.

belochub:
@aqrln let's start with rewriting existing tests from scratch to make them easier to read and update.

belochub:
Landed in ae7a557cebde953df4178d3b4ca167646e352478.

aqrln:
Labeled it `semver-major` because it isn't a backward-compatible change.

PR #72, opened 2017-02-13
lib: fix behavior with util.inspect

aqrln:
If one tried to print the result of `require('metarhia-jstp')` in general or the `Connection` class itself to console, it resulted in invocation of internal logic of the `Connection` and crashing with an exception.

A similar issue had been fixed in f7c393b0171a63b6e428b72c5ecb381635714376 and 76e4323ebb375737a777b317d250ec566692e03e but was introduced again in 4a7941acb81449c1febd51ab1d6270636e43fe99.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -13,6 +13,11 @@ module.exports = Connection;
 
 let nextConnectionId = 0;
 
+// Mapping of packet types to handler function names (forward declaration, see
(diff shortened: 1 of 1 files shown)


aqrln:
Landed in 3cec5ed7b096b0c06011c9ad8ea9a4ea6a1f7e3a.

PR #73, opened 2017-02-14
deps,build: update webpack to 2.x

aqrln:
This PR updates webpack to the next major version.

* Update the `devDependency` field in `package.json`.
* Use the new loaders config syntax.
* Drop obsolete plugins.
* Remove the options that are now enabled by default.
* Together with the previous change, remove the options that have been
  enabled by default in webpack 1.x too.


aqrln:
Landed in 5f350bcef4e6420398326fbe2430183910982256.

PR #74, opened 2017-02-14
build,test: don't recompile sources from scratch

aqrln:
`node-gyp rebuild` used to be invoked before each `npm test`. It slowed
down tests and forced CI to recompile the sources twice before doing
anything. Now `node-gyp build` is run before tests if the `build`
directory is present (otherwise it still runs `node-gyp rebuild` so
that you can do `rm -r build && npm test` if you want to).

Changes to npm scripts:

  * `npm run build-node` now runs `node-gyp build` instead of
    `node-gyp rebuild`.
  * New script `npm run rebuild-node` forces `node-gyp rebuild`.

### DIFF
--- a/build-native.js
+++ b/build-native.js
@@ -6,29 +6,36 @@ const childProcess = require('child_process');
 const EXIT_SUCCESS = 0;
 const EXIT_FAIL = 1;
 
-const nodeGyp = childProcess.spawn('node-gyp', ['rebuild'], { shell: true });
-const errorLines = [];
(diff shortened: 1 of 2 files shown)


aqrln:
@belochub I updated the short commit message so that the language isn't misleading. The full description has already been describing the fact you pointed at.

aqrln:
Landed in f0b7acad365af9265a2aed64ca507916f4ec3f61.

PR #75, opened 2017-02-20
parser: fix bug causing node to crash

belochub:
Somehow this bug was missed during previous reviews.


belochub:
Landed in 669532eefe00aa80a7b6b1bbc5c231eb850476f2.

PR #77, opened 2017-02-20
tools: automate the release preparation

aqrln:
This tool automates the release process we currently have with one neat
addition: now we'll have changelogs too, which the tool automatically
generates.

It doesn't cover the whole release process, only the most chore parts
of it: selecting the right commits and generating a changelog for them.

### DIFF
--- a/tools/prepare-release.js
+++ b/tools/prepare-release.js
@@ -206,12 +206,12 @@ function processCommits(commits) {
   let changelog = '';
 
   for (const commit of commits) {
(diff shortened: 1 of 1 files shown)


aqrln:
Generated commit stats format: https://gist.github.com/aqrln/7a9a846a42ff4a7731a5826418999112

aqrln:
Landed in 8007979cd7e5e149752b0cd7ee73c68872c1d75c.

Issue #76, opened 2017-02-20
Impossibility to catch `error` event arising in connections during handshake


belochub:
At the moment, when a server is being created and someone sends incorrect first packet (which must be handshake, but can be anything), node will crash with message "Unhandled 'error' event", even if there is an `error` event listener for this server.
That happens because we forward `error` event listener from transport to server (as `connectionError`) only after handshake was processed.
I guess possible fix will be to forward `error` event to server on connection creation.

belochub:
This bug seems to be critical, because it can cause server downtime by spamming incorrect packets on connection.

aqrln:
@belochub can you check if https://github.com/metarhia/JSTP/pull/78 fixes the issue?

PR #78, opened 2017-02-20
server: don't allow invalid packets to crash server

aqrln:
This commit fixes a critical issue that allowed to crash the server
with an unhandled error event sending an invalid packet that causes
parser error before a handshake.

Fixes: https://github.com/metarhia/JSTP/issues/76

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -311,6 +311,7 @@ Connection.prototype._onSocketClose = function() {
 //   error - error that has occured
 //
 Connection.prototype._onSocketError = function(error) {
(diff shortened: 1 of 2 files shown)


aqrln:
@belochub I've added a new commit. PTAL.

aqrln:
These shouldn't be squashed though they'll share the same 'PR-URL' field in metadata. That's okay since they are independent and logically different yet tightly related changes.

aqrln:
Landed in c211c66 and fc1d99d.

PR #84, opened 2017-02-22
socket: use socket.destroy() properly

aqrln:
`socket.destroy()` should only be called in case of application errors
that make it impossible to continue working with this socket (like
protocol parsing errors, exactly our situation). This method emits
`error` event on socket automatically if an exception is passed as the
first argument.

### DIFF
--- a/lib/transport.socket.js
+++ b/lib/transport.socket.js
@@ -208,8 +208,6 @@ SocketTransport.prototype.end = function(data) {
   } else {
     this.socket.end();
   }
-
-  this.socket.destroy();
 };
 
 // Socket data handler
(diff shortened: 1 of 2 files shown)


aqrln:
Maybe we should revert d0fb9cd053b2138a97923c605de20d3bee9ad084 after this.

belochub:
@aqrln, what about semver for this PR? Can you add a corresponding label?

aqrln:
@belochub not that it should break anything, I highly doubt that, but I feel like it's worth landing on 0.7 and leaving 0.6 as it is.

aqrln:
@belochub ping. Are there any objections to this change or I can get a sign-off and land it?

belochub:
@aqrln, I have no objections to this change, you can land it.

aqrln:
Landed in 302f352f7441d7379e2759fa51b38489b82a80b5.

PR #87, opened 2017-02-26
lint: update remark

aqrln:
Update the Markdown linting tools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,9 +49,9 @@
     "karma-safari-launcher": "^1.0.0",
     "karma-webpack": "^2.0.1",
     "mocha": "^3.2.0",
-    "remark-cli": "^2.1.0",
-    "remark-lint": "^5.4.0",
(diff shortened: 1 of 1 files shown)


tshemsedinov:
@aqrln I think such changes may pass without PR/review

aqrln:
Landed in 691795bfb0f6ff2740ebbd53abc54ccb033918b8.

aqrln:
@tshemsedinov IMO, if something is about to go to `master`, someone has to check it, even changes so little can bring something unexpected, especially since it this concrete case a dependency is updated to the next semver-major version. That's the first point that covers the necessity of review and it may be arguable. The second point is about PRs themselves: even if one doesn't wait for changes to be reviewed, a PR must be opened anyway because we have CI only build pull requests and our release preparation tooling relies on the PRs, labels in these PRs and commit metadata.

PR #96, opened 2017-03-03
connection: change style of a forward declaration

aqrln:
ESLint complains that a variable is never reassigned if it is declared
like this:

```js
let variable;
...
variable = value;
```

Because of this, we had a forward declaration in `connection.js` marked
with a directive for ESLint to ignore the line. This commit changes it
to assignment to null.


aqrln:
Landed in f6f0aba05a2e4fd15fc72adda54ede74dc918765.

PR #97, opened 2017-03-03
lib: change multiline function signatures style

aqrln:
Make multiline function signatures more concise and readable.

### DIFF
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -24,8 +24,8 @@ apps.Application = Application;
 //   args - method arguments
 //   callback - method callback
 //
-Application.prototype.callMethod = function(
-    connection, interfaceName, methodName, args, callback) {
+Application.prototype.callMethod = function(connection, interfaceName,
+    methodName, args, callback) {
   const appInterface = this.api[interfaceName];
   if (!appInterface) {
     return callback(errors.ERR_INTERFACE_NOT_FOUND);
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -25,12 +25,7 @@ apps.Application = Application;
 //   callback - method callback
 //
 Application.prototype.callMethod = function(
-  connection,
-  interfaceName,
-  methodName,
-  args,
-  callback
-) {
+    connection, interfaceName, methodName, args, callback) {
   const appInterface = this.api[interfaceName];
   if (!appInterface) {
     return callback(errors.ERR_INTERFACE_NOT_FOUND);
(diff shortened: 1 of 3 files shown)


belochub:
In my opinion, it will be better to use another style, like this one:
```javascript
Client.prototype.connectAndHandshake = function(appName, username,
    password, callback) {
```
Or something like that:
```javascript
Client.prototype.connectAndHandshake = function(appName,
                                                username,
                                                password,
                                                callback) {
```
The latter style is used in Node.js core and looks more suitable for this case.

aqrln:
@belochub I'm all for the latter style but AFAIK @tshemsedinov objects to it. Though I'm fine with the first one two.

tshemsedinov:
I like:
```js
Client.prototype.connectAndHandshake = function(
  appName, // comments
  username, // comments
  password, // comments
  callback // comments
) {
```

aqrln:
@tshemsedinov yeah, and that's what we currently have, but that style has appeared to be not very readable and making it very hard to skim through the source code when there's a lot of such functions (like what I am constantly experiencing trying to read MetaSync or GlobalStorage sources) since their bounds are not very evident for the eye (or that may be a matter of habit, IDK). Though if the parameters were indented with four spaces, I'd find it somewhat more readable since they wouldn't meld with the function body.

@belochub what do you think about it?

belochub:
@aqrln, I like the way you styled it in last commit.
@tshemsedinov, since we don't add comments on arguments like in the style you like (because we have descriptions in front of functions including detailed argument description), we may use style which @aqrln added in last commit (former of those described here https://github.com/metarhia/JSTP/pull/97#issuecomment-284267430)

tshemsedinov:
If parameters need no explanation we can just:
```js
Client.prototype.connectAndHandshake = function(
  appName, username, password, callback
) {
  // code
}
```
or comment just certain parameters:
```js
Client.prototype.connectAndHandshake = function(
  appName, username, password,
  callback // function(err, data)
) {
  // code
}
```
and make comments for function itself (not parameters), add examples, etc.
```js
Client.prototype.connectAndHandshake = function(
  // Comments for function
  appName, username, password, callback
  // Example: ...
) {
  // code
}
```
As for me, it doesn't meld with body.

aqrln:
Then my vote goes for

```
Client.prototype.connectAndHandshake = function(
  appName, username, password, callback
) {
  // code
}
```

I'm not really a fan of this style but at least I find it acceptable. So let's use it and close the PR without further discussion since the whole issue is bikeshedding while we have much more important tasks.

aqrln:
Landed in 9edd03f866bf2acdfc5b70300f97f7b6be548161.

Issue #58, opened 2017-02-06
Get rid of redundant parameters in functions


aqrln:
Some public callbacks and, maybe, core functions take redundant parameters that can be extracted from the other ones. For example, there's no need to pass an application to a function if a connection is already supplied.

PR #104, opened 2017-03-20
client: drop redundant callback argument

aqrln:
This commit removes the sessionId argument of client.connectAndHandshake() and client.connectAndInspect(). It is not needed since the connection is passed, and the session id is
available as a property of the connection.

Fixes: https://github.com/metarhia/JSTP/issues/58


aqrln:
Landed in b6e9eee0144ee03960454f769983acbcac65b90a.

PR #105, opened 2017-03-21
client: handle errors in connectAndInspect

aqrln:
As `DataCollector` always returns the data it managed to collect
successfully and passes errors as a separate object, `connectAndInspect`
has put errors as `_errors` property of the result to be consistent.
This was a horrible design decision that was counter-intuitive and would
only lead to the lack of proper error handling in application code (and
has actually led in our integration test).

Worse than that, even this `_errors` property would never have been
populated with errors because of a bug in `connectAndInspect`.  A patch
for the bug (just in case we ever want to backport it):

```diff
diff --git a/lib/client.js b/lib/client.js
index 83236d9..b0436fd 100644
--- a/lib/client.js
+++ b/lib/client.js
@@ -124,7 +124,7 @@ Client.prototype.connectAndInspect = function(
       interfaces.forEach((interfaceName) => {
         connection.inspectInterface(interfaceName, (error, appInterface) => {
           if (error) {
-            appInterface = null;
+            appInterface = error;
           }
           collector.collect(interfaceName, appInterface);
         });
```

The proper solution would have been to stop inspecting interfaces as
soon as the first error occurs and return it as the first argument of
the callback.  As it turned out, it was quite tricky to do with
DataCollector, and some other issues were found while inspecting
`metasync` sources too.  Fixing these problems requires
backwards-incompatible changes and will certainly happen in the next
semver-major release of `metasync`, but we need a quick fix here.  For
this reason, the function was refactored to use promises.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -3,8 +3,6 @@
 const events = require('events');
 const util = require('util');
 
-const metasync = require('metasync');
-
 const apps = require('./applications');
 const common = require('./common');
 const Connection = require('./connection');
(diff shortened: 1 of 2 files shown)


aqrln:
Marking it semver-major because of the changed prototype of the resulting object. I'll open a backport PR for v0.6 when this one is approved.

aqrln:
Rebased to resolve conflicts.

aqrln:
Landed in d8ba3fe7b9c1832ae955d2a0a5a87ff165b768ec.

aqrln:
Well, it can't really be converted to semver-patch since there's no `_errors` property anymore. I'll apply the patch mentioned in the commit message instead then.

aqrln:
@belochub though, given it has an underscore in its name, can we say that it's an undocumented implementation detail and the bug was that `connectAndInspect()` had no error handling at all? This way we can backport this to v0.6.

aqrln:
@belochub huh, I think we indeed can. This `_error` property could never appear because of the bug with `DataCollector` usage.

belochub:
@aqrln, yeah, you are right.

aqrln:
Backported in https://github.com/metarhia/JSTP/pull/106.

PR #110, opened 2017-04-02
(v0.5 backport) src,build: improve the native module subsystem

aqrln:
* Split `jsrs-impl.cc` into separate modules.
* Make some refactoring.
* Rename the native addon to `jstp` since there already is a function
  that is not a part of JSRS.
* Fix `binding.gyp`: make `cflags` not ignored on macOS (as it appeared
  they used to be) and do not use `-O3` in Debug configuration.
* Use a macro to throw V8 exceptions to avoid boilerplate code.

Backport-of: https://github.com/metarhia/JSTP/pull/36


belochub:
@aqrln, why didn't you delete `jsrs-impl.cc` like you did in original PR?

aqrln:
@belochub whoops, looks like I messed up the conflict resolution a bit. Thanks for catching that!

aqrln:
@belochub I have removed the obsolete file. A new commit that fixed the build is now is this PR too, so the CI is passing. Can you please take at look?

aqrln:
Landed in 0d7b66b86e7fd2c764d687a4cb46a870dd7125e3 and 64aae38b145d49efc870f66a8864de097254cb26.

PR #113, opened 2017-04-02
(v0.5 backport) connection: handle optional callbacks properly

aqrln:
For some public methods of Connection that perform network actions
callbacks are optional. This commit fixes the behavior of default
actions when a callback function is not passed so that (1) the system
logic is always run and (2) a Connection doesn't complain about
callback packets for unknown source packets.

Backport-of: https://github.com/metarhia/JSTP/pull/52

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -82,3 +82,8 @@ common.createZeroFilledBuffer = function(size) {
 if (Buffer.alloc) {
   common.createZeroFilledBuffer = Buffer.alloc;
 }
+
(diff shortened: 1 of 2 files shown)


aqrln:
Sorry, accidentally pushed to the target branch, had to force-push. Can I get a quick LGTM, btw?

aqrln:
Landed in 3d3067b204976f0d0815072f3ce66d60f21aa2a1.

PR #122, opened 2017-04-03
add arrow-parens rule to eslint

lundibundi:
Force eslint to throw an error on arrow function with braces
but without parens, this will make code a bit more consistent

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -136,6 +136,10 @@ rules:
   no-use-before-define:
     - error
     - functions: false
+  arrow-parens:
+    - error
+    - as-needed
+    - requireForBlockBody: true
   arrow-body-style:
     - error
     - as-needed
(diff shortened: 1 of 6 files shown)


aqrln:
I'd actually prefer if the parens were not necessary for one-liners, but would not be treated as errors. Seems like that's exactly the reason I did not enable this rule 😕

aqrln:
@lundibundi can you please fix the linter errors and add the "lint:" prefix to the commit message? I'd like to see how those functions will look like and how will it impact readability.

lundibundi:
I guess it's better to keep track of functions with braces but without parens, as is rare to use parens with simple functions

lundibundi:
@aqrln done

aqrln:
Landed in 29938d2e5599f5bdaf36fd93b6c3ee5745695c71.

PR #127, opened 2017-05-05
cli: move cli.js to lib

lundibundi:


lundibundi:
@aqrln oh, I got it. But to what extent should I split them, as of now I see 2 ways either move everything to lib and only expose interface with in\out streams + options (in future - use * events, connect on start, etc) or move what is know called lineProcessor entirely to bin along with readline interface and make all user input processing in bin?

aqrln:
@lundibundi I'd say everything that knows about `process.stdin`, `process.stdout`, `process.on('SIGINT', ...)` and `process.exit()` should got to `bin/cli.js`, everything else to `lib/cli.js`. It may require some refactoring though, since anything related to the logic should know nothing about the former, but as far as I see, `cli.js` is quite tightly coupled.

aqrln:
@lundibundi and if you're actually going to do some refactoring, feel free to create the `lib/cli/` directory and put multiple modules there :)

Let's count that as the first step in improving modularity and project structure. I'd definitely want to see the transports in their own directory too.

belochub:
@lundibundi ping.

lundibundi:
Also if anyone knows how to create commandProcessor and lineProcessor without the need to write all of their code in constructor or usage of utils.bindAll, I'd gladly change that as I don't like it either.

aqrln:
@lundibundi make them classes too, pass instances of `Cli` to constructors to save them as properties, and use `this.cli.whatever` in all of their methods.

bin/cli.js
```diff
@@ -0,0 +1,43 @@
...
+  rl.write('\n');
+  if (userInput) rl.write(userInput);
+};
+
+const finish = () => {
+  rl.close();
+  process.exit();
+};
+
+const cli = new Cli(log);
```
aqrln: Can we pass the whole readline interface instance here?

bin/cli.js
```diff
@@ -0,0 +1,43 @@
...
+
+'use strict';
+
+const readline = require('readline');
+
+const Cli = require('../lib/cli/cli');
+
+let rl = null;
+
+const log = (msg) => {
```
aqrln: I believe this piece of logic should belong to the `Cli` class.

bin/cli.js
```diff
@@ -0,0 +1,43 @@
...
+  const userInput = rl.line;
+  if (userInput) rl.clearLine();
+  console.log(msg);
```
aqrln: And it should write to the underlying stream directly (`process.stdout` in this case, but it may be anything that the readline interface was created on top of) instead of using `console.log()`.

bin/cli.js
```diff
@@ -0,0 +1,43 @@
...
+rl.on('line', (line) => {
+  cli.processLine(line, rl.prompt(true));
+});
+
+rl.on('close', () => finish());
+
+rl.on('SIGINT', () => finish());
+
+rl.prompt(true);
+
```
aqrln: Extraneous blank line.

lib/cli/cli.js
```diff
@@ -0,0 +1,269 @@
...
+  });
+};
+
+// Map all remaining commands directly
+Object.keys(commandProcessor).map((command) => {
+  if (!lineProcessor[command]) {
+    lineProcessor[command] = commandProcessor[command];
+  }
+});
+
```
aqrln: Extraneous blank line. I often see missing blank lines, but making two of them is something new :)

lib/cli/utils.js
```diff
@@ -0,0 +1,97 @@
...
+  // they'll be not valid (those are completions for previous inputs)
+  if (newDepth <= inputs.length - 1) completions = [];
+  if (!completions[0] ||
+      completions.length === 1 && completions[0] === inputs[depth]) {
+    completions = [];
+    if (completer.help) help = completer.help();
+  }
+  return [completions, help];
+};
+
```
aqrln: Ditto

bin/cli.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
```
aqrln: I don't see the mode of this file. Is it executable? If not, can you please `chmod` it to be?

lib/cli/cli.js
```diff
@@ -0,0 +1,269 @@
...
+module.exports = Cli;
+
+const commandProcessor = {};
+const lineProcessor = {};
+
+// log - logger function to print results
+//       help and error messages
+//
+function Cli(log) {
```
aqrln: Can you please use an ES2015 class here?

bin/cli.js
lundibundi: @aqrln sure, forgot about it.

lib/cli/cli.js
lundibundi: @aqrln sure.

bin/cli.js
lundibundi: @aqrln yeah, missed that while moving it out of cli.

bin/cli.js
lundibundi: @aqrln I'm not sure if it'll be better for abstraction as now Cli only depends on the function to output data and not on the  whole readline interface. Also there is another problem - readline needs `completer` upon its creation which is only present after Cli has been created.

bin/cli.js
aqrln: Still not executable ;)

lib/cli/cli.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const EventEmitter = require('events').EventEmitter;
+
+const CommandProcessor = require('./command-processor');
+const LineProcessor  = require('./line-processor');
+const utils = require('./utils');
+
+module.exports = class Cli extends EventEmitter {
+  // log - logger function to print results
```
belochub: Missing comma here after 'results'.

aqrln:
Fixed the permissions myself because @lundibundi's local repo occurred to be on a disk with a non-Unix filesystem which he shares between operating systems.

lib/cli/cli.js
```diff
@@ -0,0 +1,72 @@
...
+    this.lineProcessor = new LineProcessor(this.commandProcessor);
+
+    // Map all remaining commands directly
+    utils.filterKeys(
+      Object.getOwnPropertyNames(CommandProcessor.prototype),
+      ['constructor', 'complete', 'getNextCompleter']
+    ).map((command) => {
+      if (!this.lineProcessor[command]) {
+        this.lineProcessor[command] =
+          this.commandProcessor[command].bind(this.commandProcessor);
```
aqrln: One more thing: I'd find it more readable to be `(...args) => this.commandProcessor[command](...args)`, but I'm fine with it either way.

lib/cli/cli.js
```diff
@@ -0,0 +1,72 @@
...
+      return callback(null);
+    }
+
+    const cmd = utils.tryCompleter(type, this.commandProcessor);
+
+    const processor = this.lineProcessor[cmd];
+    if (!processor) {
+      this.log(`Unknown command '${cmd}'`);
+    } else {
+      processor.bind(this.lineProcessor)(leftover, (err, result) => {
```
belochub: Can you please change it to use `apply` instead of `bind` with immediate call?

lib/cli/utils.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+const utils = {};
+module.exports = utils;
+
+// Simple wrapper function to filter out elements that starts
```
belochub: Word "elements" is plural, thus it should be "start", not "starts".

aqrln:
Landed in ecdecab82ac69878aa3787d691d788bb5638dfd8.

Issue #125, opened 2017-05-05
Autocompletion in CLI


aqrln:
This is a meta-issue for two sub-tasks:

* [x] Autocompletion for CLI commands (https://github.com/metarhia/JSTP/pull/132)
* [x] Autocompletion for method names (https://github.com/metarhia/JSTP/pull/141)

PR #132, opened 2017-05-05
cli: implement autocompletion for cli commands

lundibundi:
Implements 'Autocompletion for CLI commands' and allows to implement 'Autocompletion for method names' (#125)

The behavior is as follows:
* on empty string (or trailing space) just show what commands are available for current level
* on partially entered commands show appropriate completions
* on fully entered command with no trailing space or when no completions available show help
* allow to use partially entered command without completing them it there are no collisions

### DIFF
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -24,13 +24,13 @@ const log = (msg) => {
   if (userInput) rl.write(userInput);
 };
 
-function completeByFields(input, object) {
-  return Object.keys(object)
-    .filter(c => !c.startsWith('_') && (!input || c.startsWith(input)));
+function complete(input, completions) {
+  if (!input) return completions;
+  return completions.filter(c => c.startsWith(input));
 }
 
 function tryComplete(input, completer) {
-  const completions = completer._complete([input], 0)[0];
+  const completions = completer.complete([input], 0)[0];
   if (completions.length === 1) return completions[0];
   return input;
 }
@@ -57,21 +57,21 @@ function iterativeCompletion(inputs, depth, completer) {
     const nextCompleter = completer[completions[0]];
     if (!nextCompleter) return [completions, help];
 
-    if (nextCompleter._complete && depth < inputs.length) {
-      const [newCompletions, newDepth] = nextCompleter._complete(inputs, depth);
+    if (nextCompleter.complete && depth < inputs.length) {
+      const [newCompletions, newDepth] = nextCompleter.complete(inputs, depth);
       return helper(newDepth, depth, nextCompleter, newCompletions);
     }
     if (inputs[oldDepth] === completions[0]) {
       completions.shift();
-      if (nextCompleter._help) help = nextCompleter._help();
+      if (nextCompleter.help) help = nextCompleter.help();
     }
     return [completions, help];
   }
-  if (completer._complete) {
-    const [newCompletions, newDepth] = completer._complete(inputs, depth);
+  if (completer.complete) {
+    const [newCompletions, newDepth] = completer.complete(inputs, depth);
     return helper(newDepth, depth, completer, newCompletions);
   }
-  if (completer._help) return [[], completer._help()];
+  if (completer.help) return [[], completer.help()];
   return [[], ''];
 }
 
@@ -109,9 +109,10 @@ const state = {
   connection: null
 };
 
-commandProcessor._complete = (inputs, depth) => {
+commandProcessor.complete = (inputs, depth) => {
+  const completions = ['call', 'event', 'connect', 'disconnect', 'exit'];
   const cmd = inputs[depth];
-  return [completeByFields(cmd, commandProcessor), depth + 1];
+  return [complete(cmd, completions), depth + 1];
 };
 
 commandProcessor.call = (interfaceName, methodName, args, callback) => {
@@ -119,7 +120,7 @@ commandProcessor.call = (interfaceName, methodName, args, callback) => {
   state.connection.callMethod(interfaceName, methodName, args, callback);
 };
 
-commandProcessor.call._help = () => (
+commandProcessor.call.help = () => (
   'call <interfaceName> <methodName> [ <arg> [ , ... ] ]'
 );
 
@@ -129,7 +130,7 @@ commandProcessor.event = (interfaceName, eventName, args, callback) => {
   callback();
 };
 
-commandProcessor.event._help = () => (
+commandProcessor.event.help = () => (
   'event <interfaceName> <eventName> [ <arg> [ , ... ] ]'
 );
 
@@ -148,7 +149,7 @@ commandProcessor.connect = (host, port, appName, callback) => {
   );
 };
 
-commandProcessor.connect._help = () => (
+commandProcessor.connect.help = () => (
   'connect <host>:<port> <application name>'
 );
 
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -189,7 +189,7 @@ function _split(str, separator, limit, leaveEmpty) {
   let start = 0;
 
   const shouldPush = end =>
-    start !== end || (leaveEmpty && !result[result.length - 1]);
+    start !== end || (leaveEmpty && result[result.length - 1] !== '');
 
   // eslint-disable-next-line no-unmodified-loop-condition
   while (!limit || result.length < limit) {
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -110,7 +110,7 @@ const state = {
 };
 
 commandProcessor.complete = (inputs, depth) => {
-  const completions = ['call', 'event', 'connect', 'disconnect', 'exit'];
+  const completions = ['call', 'connect', 'disconnect', 'event', 'exit'];
   const cmd = inputs[depth];
   return [complete(cmd, completions), depth + 1];
 };
@@ -174,7 +174,7 @@ commandProcessor.exit = () => {
 // limit - resulting length of output array - 1 (last one is what's left),
 //         if !limit === true => means no limit and split till no more
 //         separators found
-// leaveEmpty - if true multiple separators in sequence will be added as empty
+// leaveEmpty - if true multiple separators in sequence will be added as
 //              empty string, else they are skipped
 //
 // returns an array of strings
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -67,8 +67,12 @@ function iterativeCompletion(inputs, depth, completer) {
     }
     return [completions, help];
   }
-  const [newCompletions, newDepth] = completer._complete(inputs, depth);
-  return helper(newDepth, depth, completer, newCompletions);
+  if (completer._complete) {
+    const [newCompletions, newDepth] = completer._complete(inputs, depth);
+    return helper(newDepth, depth, completer, newCompletions);
+  }
+  if (completer._help) return [[], completer._help()];
+  return [[], ''];
 }
 
 rl.on('line', (line) => {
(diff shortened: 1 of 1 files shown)


aqrln:
@lundibundi can you please add the `Refs: https://github.com/metarhia/JSTP/issues/125` metadata field to the description of your second commit? Thanks!

tools/cli.js
```diff
@@ -101,24 +164,37 @@ commandProcessor.exit = () => {
...
 function _split(str, separator, limit, leaveEmpty) {
-  const shouldTrim = (start, split) => !leaveEmpty && start === split;
+  const isLastEmpty = arr => !arr[arr.length - 1];
 
   const result = [];
   let start = 0;
 
   // eslint-disable-next-line no-unmodified-loop-condition
-  for (let i = 0; !limit || i < limit; i++) {
+  while (!limit || limit - result.length > 0) {
```
belochub: In my opinion, rewriting second condition as
```javascript
result.length < limit
```
will make it more readable.

lundibundi:
@aqrln done

tools/cli.js
lundibundi: ok

tools/cli.js
```diff
@@ -101,24 +164,37 @@ commandProcessor.exit = () => {
...
     const split = str.indexOf(separator, start);
     if (split === -1) break;
-    if (!shouldTrim(start, split)) {
+    if (start !== split || leaveEmpty && !isLastEmpty(result)) {
```
belochub: This complex condition really needs parentheses.

tools/cli.js
```diff
@@ -24,14 +24,64 @@ const log = (msg) => {
   if (userInput) rl.write(userInput);
 };
 
+function completeByFields(input, object) {
+  return Object.keys(object)
+    .filter(c => !c.startsWith('_') && (!input || c.startsWith(input)));
```
aqrln: What type can `input` have? If it's always a string, then the `!input` check is superfluous, `c.startsWith('')` will always return `true`.

tools/cli.js
```diff
@@ -101,24 +164,37 @@ commandProcessor.exit = () => {
...
+//              empty string, else they are skipped
+//
+// returns an array of strings
+//
+// the behaviour is as follows:
+//  splits 'str' till limit is bound or no more separators left in 'str'
+//  if leaveEmpty is true then multiple separators in sequence are written in
+//  resulting array as one empty string (''), else they are skipped
+//  and doesn't get counted to limit
 function _split(str, separator, limit, leaveEmpty) {
```
aqrln: Not related to this PR, but why does this function start with an underscore? It isn't exported so it's private anyway.

tools/cli.js
```diff
@@ -101,24 +164,37 @@ commandProcessor.exit = () => {
...
-    if (!shouldTrim(start, split)) {
+    if (start !== split || leaveEmpty && !isLastEmpty(result)) {
       result.push(str.slice(start, split));
-    } else  {
-      i--;
     }
     start = split + separator.length;
   }
-  if (!shouldTrim(start, str.length)) {
+  if (start !== str.length || leaveEmpty && !isLastEmpty(result)) {
```
belochub: It is almost the same condition as on line 192, you should probably replace it with function, like it was before with `shouldTrim`.

tools/cli.js
lundibundi: @aqrln we originally wrote it like this, as it really is private, i guess i will fix this during decomposition of cli.js in https://github.com/metarhia/JSTP/pull/127

tools/cli.js
lundibundi: @belochub yeah, but i'm not sure if it's okay to use shouldTrim (name) as it will also add '' when needed.

tools/cli.js
lundibundi: @aqrln the problem is it can be undefined when you press tab with no input, not sure how to handle this better

tools/cli.js
```diff
@@ -24,14 +24,64 @@ const log = (msg) => {
   if (userInput) rl.write(userInput);
 };
 
+function completeByFields(input, object) {
+  return Object.keys(object)
+    .filter(c => !c.startsWith('_') && (!input || c.startsWith(input)));
+}
+
+function tryComplete(input, completer) {
+  const completions = completer._complete([input], 0)[0];
```
aqrln: As far as I understand the `completer` interface, `_complete` and `_help` are its public methods. Why do they start with underscores?

tools/cli.js
```diff
@@ -24,14 +24,64 @@ const log = (msg) => {
...
+      const [newCompletions, newDepth] = nextCompleter._complete(inputs, depth);
+      return helper(newDepth, depth, nextCompleter, newCompletions);
+    }
+    if (inputs[oldDepth] === completions[0]) {
+      completions.shift();
+      if (nextCompleter._help) help = nextCompleter._help();
+    }
+    return [completions, help];
+  }
+  const [newCompletions, newDepth] = completer._complete(inputs, depth);
```
aqrln: The doc comment above this function states:

> // completer - object that has '_complete(inputs, depth)' function or ._help()
> //             or neither (no completions or help available)

But if the `completer` object has no `_complete` method, this line will fail with a `TypeError`.

lundibundi: oh, forgot about first entrance, thanks for catching that.

tools/cli.js
```diff
@@ -24,14 +24,64 @@ const log = (msg) => {
...
+    if (completions.length !== 1) return [completions, help];
+    const nextCompleter = completer[completions[0]];
+    if (!nextCompleter) return [completions, help];
+
+    if (nextCompleter._complete && depth < inputs.length) {
+      const [newCompletions, newDepth] = nextCompleter._complete(inputs, depth);
+      return helper(newDepth, depth, nextCompleter, newCompletions);
+    }
+    if (inputs[oldDepth] === completions[0]) {
+      completions.shift();
```
aqrln: This statement modifies the `completions` array. Won't it have any unwanted side-effects?

tools/cli.js
aqrln: Ah, well then. If it can be `undefined`, the check is completely valid.

tools/cli.js
lundibundi: it's needed to remove completion if user has already entered full command, what sort of side-effects?

tools/cli.js
belochub: @lundibundi, I didn't say that you should leave `shouldTrim` the way it was before. I meant that you you should replace repetitive conditions with function, like it was done before with `shouldTrim` function.

tools/cli.js
aqrln: Never mind, I was thinking of some global array of completions that unexpectedly gets deteriorated as user keeps hitting Tab. That's not the case here.

tools/cli.js
aqrln: Ah, I see. It's a hack to make objects which serve as sources for completions completers themselves.

lundibundi: @aqrln hm, at first I thought to always use fields for completion, but now I see that it'd to be better to state the completions manually, then i can get rid of the underscore.

tools/cli.js
```diff
@@ -50,26 +100,35 @@ rl.on('close', () => {
...
+);
+
 commandProcessor.event = (interfaceName, eventName, args, callback) => {
   if (!state.client) return callback(new Error('Not connected'));
   state.connection.emitRemoteEvent(interfaceName, eventName, args);
   callback();
 };
 
+commandProcessor.event._help = () => (
+  'event <interfaceName> <eventName> [ <arg> [ , ... ] ]'
```
aqrln: Events in JSTP don't have a concept of arguments, like methods do. They only have a single argument that can be of any type. Is it just the help that is incorrect here or that's how it is actually implemented so that it's a bug in CLI?

tools/cli.js
```diff
@@ -101,24 +164,37 @@ commandProcessor.exit = () => {
...
+//  resulting array as one empty string (''), else they are skipped
+//  and doesn't get counted to limit
 function _split(str, separator, limit, leaveEmpty) {
-  const shouldTrim = (start, split) => !leaveEmpty && start === split;
+  const isLastEmpty = arr => !arr[arr.length - 1];
 
   const result = [];
   let start = 0;
 
   // eslint-disable-next-line no-unmodified-loop-condition
```
aqrln: Does the linter still fail without this directive with the new loop condition?

tools/cli.js
lundibundi: @belochub I know, I meant that I couldn't come up with a good name for it.

tools/cli.js
lundibundi: @aqrln yeah(

tools/cli.js
lundibundi: @aqrln not sure what you meant, the '[ <arg> [ , ... ] ]' will be passed as an array as a third argument to Connection.emitRemoteEvent

aqrln: Yeah, but it's not supposed to be an array. It can be an arbitrary JavaScript value. At least, in the current version of the protocol.

tools/cli.js
aqrln: Anyway, it's not related to the current PR if that's how it is actually implemented in the CLI. It's a matter of a separate issue and PR.

tools/cli.js
aqrln: Anyway, it's not related to the current PR if that's how it is actually implemented in the CLI. It's a matter of a separate issue and PR.

lundibundi: yeah, it is implemented like that, so we should open an issue for that.

tools/cli.js
belochub: @aqrln, but isn't it redundant? As far as I know, `String.prototype.startsWith` will return `false` when first argument is `undefined`.

belochub: @aqrln, oh, I am sorry, I see why this check is OK here.

tools/cli.js
```diff
@@ -101,26 +169,36 @@ commandProcessor.exit = () => {
   process.exit();
 };
 
+// str - inputs string
+// separator - string to use as a separator
+// limit - resulting length of output array - 1 (last one is what's left),
+//         if !limit === true => means no limit and split till no more
+//         separators found
+// leaveEmpty - if true multiple separators in sequence will be added as empty
```
belochub: Word 'empty' is repeated twice here, probably a typo.

tools/cli.js
```diff
@@ -109,17 +109,18 @@ const state = {
   connection: null
 };
 
-commandProcessor._complete = (inputs, depth) => {
+commandProcessor.complete = (inputs, depth) => {
+  const completions = ['call', 'event', 'connect', 'disconnect', 'exit'];
```
aqrln: I'd prefer these to be sorted alphabetically, though I don't feel strongly for it.

tools/cli.js
```diff
@@ -50,26 +104,36 @@ rl.on('close', () => {
...
-  return [[], line];
-}
-
 const state = {
   client: null,
   connection: null
 };
 
+commandProcessor.complete = (inputs, depth) => {
+  const completions = ['call', 'event', 'connect', 'disconnect', 'exit'];
```
aqrln: I'd prefer these to be sorted alphabetically, though I don't feel strongly for it.

tools/cli.js
lundibundi: yeah

tools/cli.js
lundibundi: ok with me

aqrln:
Landed in f11130f1bdc4f91c778376dee9372abe9bbd6aea and 8a6bc52c81756db4b9118faff624170b00b40ba9 🎉

PR #133, opened 2017-05-05
lint: add import/no-unresolved ESLint rule

aqrln:
I am planning to rearrange the structure of `lib` a little bit, so I decided to add this rule to be safe doing this. And it is a nice thing to have generally.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -2,6 +2,8 @@ env:
   node: true
   es6: true
 extends: 'eslint:recommended'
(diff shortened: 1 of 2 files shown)


aqrln:
@belochub btw, heads up, you need to do `npm install -g eslint-plugin-import` in order for it to work in Vim. The `syntastic` plugin works with the global ESLint installation.

aqrln:
Landed in 06579df.

PR #136, opened 2017-05-06
ws: fix emitting error when callback is provided

belochub:
Fix WebSocket client emitting error on connection when callback is
provided.


belochub:
Landed in https://github.com/metarhia/JSTP/commit/5b79a1eb385653bf796857fe616f4d7824baec28.

PR #137, opened 2017-05-06
cli: avoid crashing on incorrect user input

belochub:
* Avoid crashing when port has incorrect value on connection.
* Avoid crashing when host is missing on connection.


belochub:
Landed in https://github.com/metarhia/JSTP/commit/acd5174593146077cfa7ef79b73d393a62ef78df.

Issue #126, opened 2017-05-05
Support connection via different transports in CLI


aqrln:
JSTP CLI should support connecting to servers using any of the available transports (i.e., TCP and WebSocket with or without TLS).

aqrln:
@belochub I've removed @lundibundi's assignment for now since he's now working on autocompletion as far as I know. Whoever is willing to champion on this is free to self-assign.

PR #138, opened 2017-05-06
cli: support different transports for connection

belochub:
Allow users to choose protocol when connecting to the
server by providing URI scheme in front of the host.

Closes: https://github.com/metarhia/JSTP/issues/126

### DIFF
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -2,10 +2,10 @@
 
 'use strict';
 
-// TODO: support jstp://server and jstps://server
-
 const jstp = require('..');
 const readline = require('readline');
+
+const DEFAULT_SCHEME = 'jstps';
 
 const commandProcessor = {};
 const lineProcessor = {};
@@ -134,8 +134,25 @@ commandProcessor.event.help = () => (
   'event <interfaceName> <eventName> [ <arg> [ , ... ] ]'
 );
 
-commandProcessor.connect = (host, port, appName, callback) => {
-  state.client = jstp.tcp.createClient({ host, port, secure: true });
+commandProcessor.connect = (protocol, host, port, appName, callback) => {
+  let transport;
+
+  switch (protocol) {
+    case 'jstp':
+    case 'jstps':
+      transport = jstp.tcp;
+      break;
+    case 'ws':
+    case 'wss':
+      transport = jstp.ws;
+      break;
+    default:
+      return callback(new Error(`Unknown protocol '${protocol}'`));
+  }
+
+  const url = `${protocol}://${host}:${port}`;
+
+  state.client = transport.createClient(url);
   state.client.connectAndHandshake(appName, null, null,
       (err, connection) => {
         if (err) return callback(err);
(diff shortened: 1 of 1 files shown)


tools/cli.js
```diff
@@ -251,15 +272,22 @@ lineProcessor.connect = (tokens, callback) => {
     return callback(reportMissingArgument('Host'));
   }
   const args = _split(tokens, ' ', 2);
-  const [host, port] = _split(args[0], ':');
+  let [scheme, authority] = _split(args[0], '://', 1, true);
+  if (authority === undefined) {
```
lundibundi: With this command `connect ://host.com app` we get wrong error `Error occurred: Host is not provided`

belochub: @lundibundi, updated.

belochub:
@lundibundi, rebased onto master, check it out once more, please.

lundibundi:
@belochub LGTM

tools/cli.js
```diff
@@ -134,8 +134,29 @@ commandProcessor.event.help = () => (
...
+      // fall through
+    case 'jstp':
+      transport = jstp.tcp;
+      options.host = host;
+      options.port = port;
+      break;
+    case 'ws':
+    case 'wss':
+      transport = jstp.ws;
+      options.url = `${protocol}://${host}:${port}`;
```
aqrln: But the TCP transport supports the URLs too, doesn't it? We may just replace the whole `switch` statement with a template literal.

tools/cli.js
```diff
@@ -247,12 +268,19 @@ lineProcessor.event = (tokens, callback) => {
...
 
 lineProcessor.connect = (tokens, callback) => {
-  if (tokens === undefined || tokens.trim().startsWith(':')) {
+  if (tokens === undefined) {
     return callback(reportMissingArgument('Host'));
   }
   const args = _split(tokens, ' ', 2);
-  const [host, portString] = _split(args[0], ':');
-  if (portString === undefined) {
+  let [scheme, authority] = _split(args[0], '://', 1, true);
```
aqrln: Maybe using `url.parse()` would be better than parsing it manually?

belochub: Tried it at first, but it turned out to be difficult to reply with meaningful error messages when using `url.parse()`.

aqrln: Oh, okay. I'm fine with it as long as it works.

tools/cli.js
belochub: Yeah, that is right, but we will still need to check for the transport type, which means that this `switch` statement won't be completely eliminated. Instead, it can be shortened a little bit.

belochub:
@aqrln, Unix domain sockets are okay by me, but what is the syntax that should be used with them?

belochub:
@aqrln 
Also, about using "Closes:" instead of "Fixes:", I think that the latter is more suitable for issues which are labeled as "bugs", while the former is a better way to say that feature requested in corresponding issue was implemented in this PR/commit.

belochub:
@aqrln, rebased on master and changed `switch` statement, check it out.

aqrln:
@belochub 

> Unix domain sockets are okay with me, but what is the syntax that should be used with them?

idk, maybe `ipc:///tmp/filename`?

aqrln:
Landed in d1d8144.

@belochub I've changed "Closes:" to "Fixes:" for now since that's what we (I suppose, at least) have standardized upon before. We'll discuss this question with the whole project team either in chat or in person, and if there's support for that, we'll start differentiating between closing bugs and closing feature requests in commit messages.

PR #141, opened 2017-05-07
cli: add method autocompletion

lundibundi:
Allow to specify remote interfaces that will be inspected upon connection and uses resulting api to complete calls. The syntax is as follows

`connect [<protocol>://]<host>:<port> <application name>  [ <interface> [ ... ] ]`

### DIFF
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -119,14 +119,13 @@ commandProcessor.complete = (inputs, depth) => {
 
 commandProcessor.call = (interfaceName, methodName, args, callback) => {
   if (!state.client) return callback(new Error('Not connected'));
-
   state.connection.callMethod(interfaceName, methodName, args, callback);
 };
 
 commandProcessor.call.complete = (inputs, depth) => {
   if (!state.api) return [[], depth];
 
-  let iface = inputs[depth++];
+  const iface = inputs[depth++];
   let method = inputs[depth];
   // there may be multiple spaces between interface and method names
   // this function completes both of them so handle empty element ('')
@@ -136,23 +135,10 @@ commandProcessor.call.complete = (inputs, depth) => {
   }
 
   let completions = complete(iface, Object.keys(state.api));
-
-  const fullInterfaceMatch = completions.some(el => el === iface);
-
-  // we should show completions if
-  // there is more than one completion and no method provided
-  if (completions.length !== 1 && method === undefined ||
-      // there is more than one completion and neither interface fully matches
-      completions.length !== 1 && !fullInterfaceMatch ||
-      // there is no method name and neither interface fully matches
-      method === undefined && !fullInterfaceMatch) {
+  if (method === undefined || !completions.some(el => el === iface)) {
     return [completions, depth];
   }
 
-  // nothing after full interface name -> show help
-  if (method === undefined) return [[], depth];
-
-  if (!fullInterfaceMatch) iface = completions[0];
   completions = complete(method, state.api[iface]);
   if (completions.length === 1 && method === completions[0]) {
     // full method name -> show help
@@ -342,14 +328,7 @@ lineProcessor.connect = (tokens, callback) => {
   if (appName === undefined) {
     return callback(reportMissingArgument('Application name'));
   }
-  let interfaces = [];
-  if (args[2]) {
-    try {
-      interfaces = jstp.parse('[' + args[2] + ']');
-    } catch (err) {
-      return callback(err);
-    }
-  }
+  const interfaces = args[2] ? _split(args[2], ' ') : [];
   commandProcessor.connect(scheme, host, port, appName, interfaces, (err) => {
     if (err) return callback(err);
     callback(null, 'Connection established');
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -208,7 +208,7 @@ commandProcessor.connect = (
 
 commandProcessor.connect.help = () => (
   'connect [<protocol>://]<host>:<port> <application name> ' +
-  '[ <interface> [ , ... ] ]'
+  '[ <interface> [ ... ] ]'
 );
 
 commandProcessor.disconnect = (callback) => {
--- a/tools/cli.js
+++ b/tools/cli.js
@@ -29,7 +29,7 @@ function complete(input, completions) {
   return completions.filter(c => c.startsWith(input));
 }
 
-function tryComplete(input, completer) {
+function tryCompleter(input, completer) {
   const completions = completer.complete([input], 0)[0];
   if (completions.length === 1) return completions[0];
   return input;
@@ -65,6 +65,8 @@ function iterativeCompletion(inputs, depth, completer) {
       }
     } while (newDepth < inputs.length && completer.complete);
   }
+  // reset completions if we didn't reach last input because
+  // they'll be not valid (those are completions for previous inputs)
   if (newDepth <= inputs.length - 1) completions = [];
   if (!completions[0] ||
       completions.length === 1 && completions[0] === inputs[depth]) {
@@ -80,7 +82,7 @@ rl.on('line', (line) => {
     return rl.prompt(true);
   }
 
-  const cmd = tryComplete(type, commandProcessor);
+  const cmd = tryCompleter(type, commandProcessor);
 
   const processor = lineProcessor[cmd];
   if (!processor) {
@@ -105,7 +107,8 @@ rl.on('close', () => {
 
 const state = {
   client: null,
-  connection: null
+  connection: null,
+  api: {}
 };
 
 commandProcessor.complete = (inputs, depth) => {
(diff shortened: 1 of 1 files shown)


tools/cli.js
```diff
@@ -24,9 +24,17 @@ const log = (msg) => {
...
-  if (!input) return completions;
-  return completions.filter(c => c.startsWith(input));
+function complete(input, completions, forbidden = []) {
+  if (!forbidden && !input) return completions;
+  return completions.filter(
+    c => !startsWithAny(c, forbidden) && (!input || c.startsWith(input)));
+}
+
+function startsWithAny(str, array) {
+  for (const el of array) {
```
aqrln: I'd just do `array.some(el => str.startsWith(el))`, but this imperative approach is fine too ;)

lundibundi: Oh, this is good, will change it in a minute.

tools/cli.js
```diff
@@ -24,9 +24,17 @@ const log = (msg) => {
   if (userInput) rl.write(userInput);
 };
 
-function complete(input, completions) {
-  if (!input) return completions;
-  return completions.filter(c => c.startsWith(input));
+function complete(input, completions, forbidden = []) {
+  if (!forbidden && !input) return completions;
+  return completions.filter(
```
belochub: As far as I understand, the result of this `filter` will always have finite number of possible values on every successful connection with interface inspection, and it does not depend on any user input after connection, which means it can be optimized by computing possible values on connection establishment (after inspection) and using them afterwards.

tools/cli.js
```diff
@@ -141,7 +160,19 @@ commandProcessor.event.help = () => (
   'event <interfaceName> <eventName> [ <arg> [ , ... ] ]'
 );
 
-commandProcessor.connect = (protocol, host, port, appName, callback) => {
+function filterApiCompletions(rawApi) {
+  const api = {};
+  const forbidden = ['_', 'domain'];
+  Object.keys(rawApi).map((int) => {
```
aqrln: Can you please use `forEach` here so that the intention is clear?

aqrln:
@belochub does this LGTY after the last update?

tools/cli.js
```diff
@@ -29,7 +29,14 @@ function complete(input, completions) {
...
 
-function tryComplete(input, completer) {
+function tryComplete(input, completions) {
+  if (!completions) return input;
+  completions = complete(input, completions);
+  if (completions.length === 1) return completions[0];
+  return input;
+}
+
+function tryCompleter(input, completer) {
```
aqrln: I find it very confusing to have two functions that are completely distinct yet have names that resemble each other so close.

tools/cli.js
```diff
@@ -116,9 +124,41 @@ commandProcessor.complete = (inputs, depth) => {
 
 commandProcessor.call = (interfaceName, methodName, args, callback) => {
   if (!state.client) return callback(new Error('Not connected'));
+
+  interfaceName = tryComplete(interfaceName, Object.keys(state.api));
```
aqrln: Do we really need to complete this?

I do see a value in allowing users to type commands incompletely, that's both convenient and consistent with tools like vi. But completing method names upon Enter without requiring the user to press Tab may lead to confusing mistakes and running random commands inadvertently. I'd prefer to force users to be explicit here.

tools/cli.js
```diff
@@ -116,9 +124,41 @@ commandProcessor.complete = (inputs, depth) => {
...
+commandProcessor.call.complete = (inputs, depth) => {
+  if (!state.api) return [[], depth];
+
+  const iface = inputs[depth++];
+  let method = inputs[depth];
+  if (method === '' && inputs[depth + 1] !== undefined) {
+    method = inputs[++depth];
+  }
+
+    // (a and b and c) or (!a and (b or c))
```
aqrln: This should probably be removed.

tools/cli.js
```diff
@@ -116,9 +124,41 @@ commandProcessor.complete = (inputs, depth) => {
...
+  let completions = complete(iface, Object.keys(state.api));
+  if (completions.length !== 1 && method === undefined ||
```
aqrln: Can we have a comment explaining the logic behind these conditions above the `if` statement? Aside, how about slitting it into three `if`s with a duplicate body (it's short enough to do that). I'd find it more readable.

tools/cli.js
```diff
@@ -116,9 +124,41 @@ commandProcessor.complete = (inputs, depth) => {
...
+      completions.length !== 1 && completions[0] !== iface ||
+      method === undefined && completions[0] !== iface) {
+    return [completions, depth];
+  }
+
+  // nothing after full interface name -> show help
```
aqrln: Yeah, I was speaking about comments like this one :)

tools/cli.js
```diff
@@ -133,7 +173,19 @@ commandProcessor.event.help = () => (
   'event <interfaceName> <eventName> [ <arg> [ , ... ] ]'
 );
 
-commandProcessor.connect = (protocol, host, port, appName, callback) => {
+function filterApiCompletions(rawApi) {
+  const api = {};
+  const forbidden = ['_', 'domain'];
+  Object.keys(rawApi).forEach((int) => {
+    api[int] = Object.keys(
+      rawApi[int]).filter(c => !forbidden.some(el => c.startsWith(el)));
```
aqrln: s/`!forbidden.some`/`forbidden.all` and invert the condition inside the lambda?

tools/cli.js
lundibundi: @aqrln yeah, I totally agree on this one but I couldn't come up with a better name, what's your opinion on `tryToComplete` for the previous one and this one to remain `tryCompleter`?

tools/cli.js
lundibundi: @aqrln actually that may really be the case, but I'd like to at least add cli option like
`-ac` / `--advancedCompletion` that'll enable this behavior as this may be convenient for advanced users. But I guess we will better make it in different PR and for now either add hard-coded config variable or comment it out comepletely untill we add the option? @belochub?

tools/cli.js
lundibundi: @aqrln surely, thanks for catching that.

tools/cli.js
lundibundi: @aqrln surely, I forgot about this one and it is quite complex.

tools/cli.js
lundibundi: @aqrln fine with me.

tools/cli.js
aqrln: @lundibundi I'd say I'm -1 on it one way or another, but I will not block the PR if I am in the minority.

tools/cli.js
aqrln: Oh, yes, I did. Thanks for correcting me. It's been about 1.5 years since I switched to Node.js from Python and I still consistently write `all` instead of `every` and `any` instead of `some` in JavaScript 😄 This works the other way around too, though. I tend to be `push`ing elements to Python lists instead of `append`ing them.

tools/cli.js
aqrln: It's more of a generic question. As far as I can see, there are functions named `complete()` and `completer()` in master already, so I'd say go ahead with `tryCompleter()` in this PR for the sake of consistency. We'll need to come up with better names for all four functions in another PR though.

tools/cli.js
lundibundi: btw, ain't `all` and `any` feels better, they are a bit more clear to me?

lundibundi:
@aqrln @belochub ping.

aqrln:
@lundibundi the CI's failing with a linter error:

```
/home/travis/build/metarhia/JSTP/tools/cli.js
  32:10  error  'tryComplete' is defined but never used  no-unused-vars
✖ 1 problem (1 error, 0 warnings)
```

tools/cli.js
```diff
@@ -116,9 +119,48 @@ commandProcessor.complete = (inputs, depth) => {
 
 commandProcessor.call = (interfaceName, methodName, args, callback) => {
   if (!state.client) return callback(new Error('Not connected'));
+
```
aqrln: Unrelated style change?

tools/cli.js
```diff
@@ -116,9 +119,48 @@ commandProcessor.complete = (inputs, depth) => {
...
 
+commandProcessor.call.complete = (inputs, depth) => {
+  if (!state.api) return [[], depth];
+
+  let iface = inputs[depth++];
+  let method = inputs[depth];
+  // there may be multiple spaces between interface and method names
+  // this function completes both of them so handle empty element ('')
+  // in between (just skip it)
+  if (method === '' && inputs[depth + 1] !== undefined) {
```
aqrln: Should this be `while`?

tools/cli.js
```diff
@@ -116,9 +119,48 @@ commandProcessor.complete = (inputs, depth) => {
...
+    method = inputs[++depth];
+  }
+
+  let completions = complete(iface, Object.keys(state.api));
+
+  const fullInterfaceMatch = completions.some(el => el === iface);
+
+  // we should show completions if
+  // there is more than one completion and no method provided
+  if (completions.length !== 1 && method === undefined ||
```
aqrln: Can `completions` be an empty array? If yes, then this condition contradicts the comment above.

tools/cli.js
```diff
@@ -286,7 +342,15 @@ lineProcessor.connect = (tokens, callback) => {
   if (appName === undefined) {
     return callback(reportMissingArgument('Application name'));
   }
-  commandProcessor.connect(scheme, host, port, appName, (err) => {
+  let interfaces = [];
+  if (args[2]) {
+    try {
+      interfaces = jstp.parse('[' + args[2] + ']');
```
aqrln: I find it a bit odd to be forced to put quotes around interface names in the `connect` command while I don't have to do that for an application name.

tools/cli.js
lundibundi: @aqrln yeah, missed it while deleting call completion.

tools/cli.js
lundibundi: @aqrln no, `_split` function collapses all consequent separators.

lundibundi:
Landed in b206c0540384b43d388538960fdff5c8e894d67c.

PR #144, opened 2017-05-07
parser: fix '[' being parsed as an empty array

belochub:


belochub:
Landed in https://github.com/metarhia/JSTP/commit/98fe33c72dfefb609930f21dea1844dd40abbf53.

PR #147, opened 2017-05-08
connection: fix error handling in optional cbs

aqrln:
This commit makes error handling in methods that take optional
callbacks more robust.

* Fix a possible TypeError in `inspectInterface()`.
* Ensure that an error is not lost in `inspectInterface()` and
  `callMethod()` by emitting it on the connection if there's no
  callback.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -77,7 +77,9 @@ Connection.prototype.callMethod = function(
 ) {
   const packet = this.createPacket('call', interfaceName, methodName, args);
   const packetId = packet.call[0];
-  this._callbacks[packetId] = callback || common.doNothing;
+  this._callbacks[packetId] = callback || ((error) => {
+    if (error) this.emit('error', error);
+  });
   this._send(packet);
 };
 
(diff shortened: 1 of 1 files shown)


aqrln:
ping @metarhia/jstp

lib/connection.js
```diff
@@ -77,7 +77,9 @@ Connection.prototype.callMethod = function(
 ) {
   const packet = this.createPacket('call', interfaceName, methodName, args);
   const packetId = packet.call[0];
-  this._callbacks[packetId] = callback || common.doNothing;
+  this._callbacks[packetId] = callback || ((error) => {
+    this.emit('error', error);
```
belochub: With this implementation it looks like that whenever someone doesn't provide callback, it always emits error on callback call. Maybe you should check if `error` is not `undefined` or `null` and emit it only in that case.
Or did I get this wrong?

aqrln: You didn't, thanks for catching that!

aqrln:
@belochub I've updated the PR, PTAL.

aqrln:
Defensively marking this as `semver-major` since it will make the Node process abort in some cases where it didn't use to. I'll open a less-aggressive backport PR which only fixes the possible `TypeError`.

aqrln:
Landed in e443548733692d110558ce85db5b99c8f002e605.

PR #151, opened 2017-05-10
Added link to  Visual Studio Code JSTP plugin

ernado-x:


aqrln:
Thanks for the contribution! I'm not sure about the need for this plugin though. As far as I can see, it adds syntax highlighting for files with extensions like `.jstp`, `.jsrs` and stuff, but we decided to drop them and use just plain `.js` almost a year ago, and now we are directing towards making our data format fully compliant to the JSON5 specs, so `.json5` will be used as the extension.

ernado-x:
It is very good that you moving in global standards direction! But in this case, you have very outdated documentations here - https://metarhia.github.io/JSTP/data-formats/ and old examples here - https://github.com/metarhia/JSTP/tree/master/examples/data-formats. It very confused me as developer which want to investigate your protocol.

aqrln:
@ernado-x yeah, the lack of up-to-date documentation is definitely an issue :)
It is one of our priorities too, right after writing decent tests and increasing the test coverage to an acceptable value, which we're working on right now. One may expect us to start writing docs and examples in a week or two.

aqrln:
/cc @metarhia/jstp

I'd want to hear some input from other collaborators about whether this is something that we want. I'd say no and go ahead saying thanks again and closing the PR, but maybe it's just me.

aqrln:
Closing the PR based on the conversation above and thumb-up votes. Please let me know if you'd like to contribute to the project in any other ways. Thanks!

Issue #145, opened 2017-05-08
Migrate existing tests to node-tap


aqrln:
The consensus is we should use `node-tap` as the test framework instead of `mocha` and `chai`. This may be a gradual and incremental process.
We'll need to migrate:
* [x] serde (#153)
* [x] message parser (#168)
* [x] errors (#167)
* [x] connection (#183, #198, #199, #200)
* [x] remote proxy (#198, #199)
* [x] client (#183, #198)
* [x] server (#183)

nechaido:
We'll need to migrate:
* [x] serde (#153)
* [ ] message parser
* [ ] errors
* [ ] connection
* [ ] remote proxy
* [ ] client
* [ ] server

PR #153, opened 2017-05-10
test: migrate serde test from mocha to tap

nechaido:
Refs: https://github.com/metarhia/JSTP/issues/145

* separate test cases from tests
* port existing serde tests to tap
* remove old mocha test

### DIFF
--- a/test/fixtures/testCases/serde.js
+++ b/test/fixtures/testCases/serde.js
@@ -1,0 +1,141 @@
+'use strict';
+
+module.exports = [
+  {
+    name: 'true',
+    value: true,
+    serialized: 'true'
+  },
+  {
+    name: 'false',
+    value: false,
+    serialized: 'false'
+  },
+  {
+    name: 'integer',
+    value: 42,
+    serialized: '42'
+  },
+  {
+    name: 'negative integer',
+    value: -3,
+    serialized: '-3'
+  },
+  {
+    name: 'float',
+    value: 1e100,
+    serialized: '1e+100'
+  },
+  {
+    name: 'float with comma',
+    value: 3.14,
+    serialized: '3.14'
+  },
+  {
+    name: 'smallFloat',
+    value: 1e-3,
+    serialized: '0.001'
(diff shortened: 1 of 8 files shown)


nechaido:
Landed in 0ded6265f3a2ed2358108e59fcd6204a293af72a

test/fixtures/testCases/serde.js
```diff
@@ -0,0 +1,141 @@
...
+    value: 1e100,
+    serialized: '1e+100'
+  },
+  {
+    name: 'float with comma',
+    value: 3.14,
+    serialized: '3.14'
+  },
+  {
+    name: 'smallFloat',
```
belochub: Style of this name is inconsistent with overall test naming style in this tests, it should be `'small float'`.

test/fixtures/testCases/serde.js
```diff
@@ -0,0 +1,141 @@
...
+    value: 'first\nsecond',
+    serialized: '\'first\\nsecond\''
+  },
+  {
+    name: 'string with single quote',
+    value: 'it\'s',
+    serialized: '\'it\\\'s\''
+  },
+  {
+    name: 'string with \\u',
```
belochub: It'd be better to say `'string with Unicode escape sequences'` here.

belochub:
I am not really happy with the fact, that you landed what seems to be a not completely trivial PR in less then an hour and without getting at least 2 reviews. But that is not the main problem here.
The real issue is the fact that we don't have any rules concerning contributions and landing.
We should probably discuss this topic and create a document. It can even be shared between some of our projects in Metarhia afterwards, especially considering the fact that there will be more people joining us in development of our technological stack in the future.

aqrln:
@belochub for the record, I told @nechaido not to land non-trivial PRs so fast right after he had done that, but +1 for developing strict guidelines and documenting the process.

aqrln:
@nechaido can you please address @belochub's feedback in a follow-up PR?

PR #158, opened 2017-05-11
test: add JSON5 specs test suite

aqrln:
This commit adds and adapts the official JSON5 parser tests.
`test/todo/json5.js` should be moved to `test/node` as soon as we fix
all the failures.


aqrln:
@metarhia/jstp-core I am going to go ahead and land this by the end of the day but I'd like to get at least two reviews.

test/todo/json5.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const tap = require('tap');
+const jstp = require('../..');
+
+const supportedByUs = {
```
belochub: We should add noctal number tests here, because we decided before that we support them.

aqrln:
@belochub @lundibundi I've added noctal tests.

aqrln:
Landed in e9db526d5cfe0bed4bbd97c9a8df0f3b875b04fc. Thanks for the reviews!

PR #159, opened 2017-05-11
npm: don't publish tests in the npm package

aqrln:
Given that https://github.com/metarhia/JSTP/pull/158 has introduced
quite a massive chunk of test fixtures, it is better to avoid including
tests with the npm package.  At this point it saves 632 KB of disk
space.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -23,3 +23,6 @@ mkdocs.yml
 
 # Examples
 examples/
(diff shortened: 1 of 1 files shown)


belochub:
Landed in d333120120e9ca29f5c29e9559179274edb45e1d.

PR #161, opened 2017-05-11
test: fix naming issues in serde test cases

nechaido:
There were issues reported in discussion of https://github.com/metarhia/JSTP/pull/153 after it was merged. This PR aims to fix them along with one renaming. Changes:
* Naming of some serde test cases
* Naming of directory for serve test cases


test/node/serde.js
```diff
@@ -4,7 +4,7 @@ const tap = require('tap');
 
 const jstp = require('../..');
 
-const testCases = require('../fixtures/test-cases');
+const testCases = require('../fixtures/serde.test-cases');
```
aqrln: I'd prefer it to be `serde-test-cases`.

nechaido:
Landed in cf4e5c4040d921fb0ac198ceb210f930f6752a4f

aqrln:
@nechaido nits generally should be addressed too, especially when they are upvoted by other collaborators, they just don't block the PR and can be fixed on landing. Aside, we have agreed upon minimal waiting time of one working day for non-trivial PRs yesterday, haven't we?

test/node/serde.js
nechaido: I don't want this to be `serde-test-cases`, because `serde` is like a namespace and it shouldn't be separated the same character used for separating words, IMHO.

nechaido:
@aqrln I'm truly sorry for missing your comment, it was not ignored I simply missed it.
As for our agreement, we also agreed that it is possible to merge PR earlier if it was approved by all maintainers. As for that - I asked @belochub in personal conversation about this, and I though I got one.
In the end, I'm deeply sorry for this whole situation. Although I'd like not to rename this directory anymore due to the fact we are going to move serde to a separate package anyway.

aqrln:
> ... due to the fact we are going to move serde to a separate package anyway

We aren't yet. It was just an idea about what we might want to do in the future.

nechaido:
@aqrln 
> And even if we did, shouldn't the project structure in that package reflect what all the collaborators think it should look like?

Yep, it should, but after moving it to another package there will be no need for `serde` prefix. That is the only reason I wrote:
> ... due to the fact we are going to move serde to a separate package anyway

aqrln:
@nechaido FWIW, the transition to a separate package isn't happening in the foreseeable future.

PR #167, opened 2017-05-15
test: migrate RemoteError tests to tap

nechaido:
Refs: #145 

* move test cases to fixtures from tests
* port RemoteError test to tap
* remove mocha tests for RemoteError


nechaido:
@metarhia/jstp-core ping

nechaido:
@metarhia/jstp-core ping

test/fixtures/remote-error-test-cases/default-messages.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const jstp = require('../../../');
+
+module.exports = Object.keys(jstp)
+  .filter(key => key.startsWith('ERR_'))
```
aqrln: These should be hardcoded instead of dynamically introspected.

test/fixtures/remote-error-test-cases/get-jstp-array.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
```
aqrln: I don't actually see any reason for the whole `fixtures/remote-error-test-cases` directory to exist. These files are not pure data, they contain quite a bit of imperative logic, and they are very small and isolated. I'd say their contents should be simply moved to the corresponding tests.

nechaido: @aqrln Thank you for your review.
I have added changes due to it, although I do not see a lot of imperative logic:
* scanning for default error messages
* calling a `RemoteError` and `TypeError` constructor

Both moved to a related test.

test/fixtures/remote-error-test-cases/default-messages.js
nechaido: @aqrln this test is for checking that every jstp error has a default message. Thus introspection is needed here in case someone adds new jstp error but forgets to add a default message.

nechaido:
@aqrln ping

nechaido:
Landed in 5ea6001ca37c34225904720a975b580c95243cce.

PR #171, opened 2017-05-18
parser: fix not throwing error in some cases

belochub:
Fix not throwing error when parsing arrays and objects that
have no closing brackets and objects that have empty values
for some of the keys.

Examples of such cases:
- `'[42,'` was parsed as `[ 42 ]`.
- `'{key:'`, `'{'` and `'{key:,}'` were parsed as `{}`.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -735,7 +735,7 @@ Local<Value> ParseArray(Isolate*    isolate,
   }
 
   if (!has_ended) {
-    THROW_EXCEPTION(SyntaxError, "Missing closing brackets in array");
+    THROW_EXCEPTION(SyntaxError, "Missing closing bracket in array");
     return Undefined(isolate);
   }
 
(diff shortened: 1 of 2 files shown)


src/jsrs_parser.cc
```diff
@@ -719,8 +734,8 @@ Local<Value> ParseArray(Isolate*    isolate,
     }
   }
 
-  if (is_empty) {
-    THROW_EXCEPTION(SyntaxError, "Invalid format in array");
+  if (!has_ended) {
+    THROW_EXCEPTION(SyntaxError, "Missing closing brackets in array");
```
aqrln: I suppose "bracket" should be singular there, shouldn't it?

belochub:
@aqrln, I fixed the typo you mentioned and added some test cases for this fix, PTAL.

belochub:
Landed in https://github.com/metarhia/JSTP/commit/27675537261f88b8a9a9d488facae5f515f3007e.

PR #176, opened 2017-05-24
test: migrate common.js tests to tap

nechaido:
Also add sinon to dependencies.

### DIFF
--- a/test/node/common-forward-event.js
+++ b/test/node/common-forward-event.js
@@ -1,61 +1,59 @@
 'use strict';
 
-const tap = require('tap');
-const sinon = require('sinon');
+const test = require('tap').test;
 
 const events = require('events');
 
 const common = require('../../lib/common');
 
-tap.test('must forward a single event', (test) => {
+test('must forward a single event', (test) => {
+  test.plan(1);
+
   const sourceEventEmitter = new events.EventEmitter();
   const targetEventEmitter = new events.EventEmitter();
-
-  const spy = sinon.spy();
 
   common.forwardEvent(sourceEventEmitter, targetEventEmitter, 'testEvent');
-  targetEventEmitter.on('testEvent', spy);
+  targetEventEmitter.on('testEvent', () => {
+    test.pass('event handler must be called');
+  });
 
   sourceEventEmitter.emit('testEvent');
-  test.assert(spy.called, 'event handler must be called');
-  test.end();
 });
 
-tap.test('must forward a single event under a new name', (test) => {
+test('must forward a single event under a new name', (test) => {
+  test.plan(1);
+
   const sourceEventEmitter = new events.EventEmitter();
   const targetEventEmitter = new events.EventEmitter();
-
-  const spy = sinon.spy();
 
   common.forwardEvent(sourceEventEmitter, targetEventEmitter,
                         'testEvent', 'renamedEvent');
-  targetEventEmitter.on('renamedEvent', spy);
+
+  targetEventEmitter.on('renamedEvent', () => {
+    test.pass('event handler must be called');
+  });
 
   sourceEventEmitter.emit('testEvent');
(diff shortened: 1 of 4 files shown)


test/node/common-safe-require.js
```diff
@@ -0,0 +1,13 @@
...
+
+const common = require('../../lib/common');
+
+const existingModule = 'tap';
+const nonExistingModule = 'tp';
+
+tap.ok(common.safeRequire(existingModule), 'must require existing module');
+
+tap.equal(common.safeRequire(nonExistingModule), null,
+  'must return \'null\' if module doesn\'t exist');
```
belochub: It will be better to avoid escaping by using backticks or double quotes here, but this is not mandatory.

nechaido: @belochub currently eslint forbids to do so.

tshemsedinov:
@metarhia/jstp-core I think we may merge `/jstp/lib/common.js` into `metarhia/Common`

test/node/common-forward-event.js
```diff
@@ -0,0 +1,61 @@
...
+  const targetEventEmitter = new events.EventEmitter();
+
+  const spy = sinon.spy();
+
+  common.forwardEvent(sourceEventEmitter, targetEventEmitter,
+                        'testEvent', 'renamedEvent');
+  targetEventEmitter.on('renamedEvent', spy);
+
+  sourceEventEmitter.emit('testEvent');
+  test.assert(spy.called, 'event handler must be called');
```
lundibundi: Ain't it 'must have been called'?

nechaido: @lundibundi yep, I'l fix it

aqrln: @lundibundi it should have been if it was a sentence, but in this case it's just a test name. Look at the other assertions, they are written in the same style.

test/node/common-forward-event.js
```diff
@@ -0,0 +1,59 @@
...
+});
+
+test('must forward a single event under a new name', (test) => {
+  test.plan(1);
+
+  const sourceEventEmitter = new events.EventEmitter();
+  const targetEventEmitter = new events.EventEmitter();
+
+  common.forwardEvent(sourceEventEmitter, targetEventEmitter,
+                        'testEvent', 'renamedEvent');
```
aqrln: The indentation os off here.

test/node/common-safe-require.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const tap = require('tap');
+
+const common = require('../../lib/common');
+
+const existingModule = 'tap';
```
aqrln: Maybe it would be better to require a core module here, but this works for me too.

test/node/common-safe-require.js
```diff
@@ -0,0 +1,13 @@
...
+const nonExistingModule = 'tp';
```
aqrln: And it might be better to use something more explicitly non-existent like `__not_a_module` here.

test/node/common-safe-require.js
```diff
@@ -4,8 +4,8 @@ const tap = require('tap');
 
 const common = require('../../lib/common');
 
-const existingModule = 'tap';
-const nonExistingModule = 'tp';
+const existingModule = '../..';
+const nonExistingModule = '__non_existinting_module__';
```
aqrln: Typo: s/existinting/existing

aqrln:
@nechaido CI's failing:

```
Error: Cannot find module '../..'
```

nechaido:
@aqrln yea, I see it. Unfortunately I had no time to fix it until now.

nechaido:
@aqrln fixed.

nechaido:
Landed in a842a5fcf82cdd6c92894fb5cb17fbdb84814b7d.

PR #177, opened 2017-05-24
doc: update links to JSTP repos

aqrln:
Update the links that became obsolete after the repo became lowercased
and the links to the Java and Swift repos that were moved to a new home.


aqrln:
Landed in 8d3d15bc2d062d133c8b5ca3f8dab58f01624bfb.

PR #178, opened 2017-05-24
parser: rework error handling

belochub:
Avoid using TryCatch introduced in
29b3e197ae4c33d1b9057b57731b4c422de764f3 by using MaybeLocal instead.

Refs: https://github.com/metarhia/jstp/pull/143

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -32,13 +32,13 @@ using v8::Integer;
 using v8::Isolate;
 using v8::Local;
 using v8::Maybe;
+using v8::MaybeLocal;
 using v8::NewStringType;
 using v8::Null;
 using v8::Number;
 using v8::Object;
 using v8::String;
 using v8::True;
-using v8::TryCatch;
 using v8::Undefined;
 using v8::Value;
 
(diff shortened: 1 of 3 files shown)


src/jsrs_parser.cc
```diff
@@ -273,7 +269,7 @@ Local<Value> ParseBool(Isolate*    isolate,
     *size = 5;
   } else {
     THROW_EXCEPTION(TypeError, "Invalid format: expected boolean");
-    result = Undefined(isolate);
+    result = MaybeLocal<Value>();
```
aqrln: Isn't this the very same value that `result` was initialized with at line 263?

belochub: Yeah, right, I will remove redundant assignments, thanks for catching that 😃

src/jsrs_parser.cc
```diff
@@ -90,16 +90,11 @@ Local<Value> Parse(Isolate* isolate, const String::Utf8Value& in) {
...
-    TryCatch trycatch(isolate);
-    result =
-        (kParseFunctions[type])(isolate, str + start_pos, end, &parsed_size);
-
-    if (trycatch.HasCaught()) {
-      trycatch.ReThrow();
-      return Undefined(isolate);
-    }
+  MaybeLocal<Value> result;
+  result = (kParseFunctions[type])(isolate, str + start_pos, end, &parsed_size);
```
lundibundi: Just a nit but maybe it's better to make it into one statement? =)

src/jsrs_parser.cc
```diff
@@ -698,20 +684,17 @@ Local<Value> ParseArray(Isolate*    isolate,
...
-        t = kParseFunctions[current_type](isolate,
-                                          begin + i,
-                                          end,
-                                          &current_length);
-        if (trycatch.HasCaught()) {
-          trycatch.ReThrow();
-          return Undefined(isolate);
-        }
+      MaybeLocal<Value> t;
+      t = kParseFunctions[current_type](isolate,
```
lundibundi: Same as before, this one will event fit actually.

aqrln:
Landed in a67612528d832ea9349e36093a6d3a941caaae22. Thanks!

Issue #146, opened 2017-05-08
Make tests that create servers listen on random ports


aqrln:
Now the ports are hardcoded so the tests fail when another server is open on the same port (which is quite common for port 3000).

nechaido:
Closed by: #183, #198, #199, #200

aqrln:
Reopening, `node/basic-tls-server` still has a hardcoded port.

Issue #53, opened 2017-02-06
Rewrite too implementation-specific Connection tests


aqrln:
They dig into implementation details making it hard to change the internal behavior.

nechaido:
Closed by: #183, #198, #199, #200.

PR #183, opened 2017-05-26
test: migrate handshake and server tests

nechaido:
Refs: #145
Refs: #146 
Refs: #53.


nechaido:
@metarhia/jstp-core ping.

nechaido:
@metarhia/jstp-core ping.

test/node/connection-handshake.js
```diff
@@ -0,0 +1,112 @@
...
+
+test('must not perform a handshake with invalid credentials', (test) => {
+  test.plan(2);
+
+  const port = 3335;
+  const server = jstp.tcp.createServer(port, [application], app.authCallback);
+  server.listen(() => {
+    const client = jstp.tcp.createClient({ host: 'localhost', port });
+
+    client.connectAndHandshake(app.name, app.login, 'psssword', (error) => {
```
lundibundi: I think it's better to change 'psssword' to something more explicitly wrong, as now it is very similar to the correct one.

test/node/connection-handshake.js
```diff
@@ -0,0 +1,112 @@
...
+
+test('must handle inexistent application error', (test) => {
+  test.plan(2);
+
+  const port = 3336;
+  const server = jstp.tcp.createServer(port, [application], app.authCallback);
+  server.listen(() => {
+    const client = jstp.tcp.createClient({ host: 'localhost', port });
+
+    client.connectAndHandshake('inexistentApp', null, null, (error) => {
```
lundibundi: maybe it'll be better as s/inexistentApp/nonexistentApp ?

nechaido:
@lundibundi fixed.

test/fixtures/application.js
```diff
@@ -0,0 +1,53 @@
...
+
+  let username = null;
+  let success = false;
+
+  if (strategy === 'anonymous') {
+    success = true;
+  }
+
+  if (strategy === 'login' &&
+    credentials[0] === login &&
```
aqrln: This line must be indented two characters more to the right.

test/fixtures/application.js
```diff
@@ -0,0 +1,53 @@
...
+    credentials[1] === password) {
```
aqrln: Ditto.

test/node/connection-handshake.js
```diff
@@ -0,0 +1,96 @@
...
+const app = require('../fixtures/application');
+const application = new jstp.Application(app.name, app.interfaces);
+
+const Transport = require('../unit/mock/transport');
+
+let server;
+let client;
+
+test.beforeEach((done) => {
+  const port = 3333;
```
aqrln: This is prone to a race condition. The server isn't guaranteed to be shut down before next test starts. However, it is unlikely to happen because of small number of connections, so let's fix it after https://github.com/metarhia/jstp/pull/190 lands.

nechaido:
Blocked until #190.

test/fixtures/application.js
nechaido: @aqrln ok, but I'd suggest to add eslint rule bohj gor this and function calls.

nechaido: @aqrln eslint expects it to be indented with 4 spaces not 6

test/fixtures/application.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const jstp = require('../../');
```
belochub: May you please be consistent in using trailing slash in the package `require`s?

nechaido: I forgot to remote them here, thank you

test/fixtures/application.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  let username = null;
+  let success = false;
+
+  if (strategy === 'anonymous') {
+    success = true;
+  }
+
+  if (strategy === 'login' &&
```
belochub: In my opinion, it could be better to merge this condition with the previous one using `else if`. It can also probably fix the indentation issue mentioned in @aqrln's comment.

test/fixtures/application.js
```diff
@@ -0,0 +1,47 @@
...
+    credentials[0] === login &&
+    credentials[1] === password) {
+    success = true;
+    username = login;
+  }
+
+  if (success) {
```
belochub: In my opinion this part would look better written like this:
```
if (!success) {
  return callback(new jstp.RemoteError(jstp.ERR_AUTH_FAILED));
}
callback(null, username, sessionId);
```
Just to remove extra indentation and to be more consistent with the style usually used in Node.js.

test/fixtures/application.js
nechaido: I don't like the way it will look:
```
if (strategy === 'anonymous') {
  success = true;
} else if (strategy === 'login' &&
           credentials[0] === login &&
           credentials[1] === password) {
  success = true;
  username = login;
}
```

nechaido: Although I don't like this either:
```
if (strategy === 'anonymous') {
  success = true;
}

if (strategy === 'login' &&
    credentials[0] === login &&
    credentials[1] === password) {
  success = true;
  username = login;
}
```
@metarhia/jstp-core could you please provide your feedback on this?

aqrln: @nechaido both of these look good to me.

nechaido:
@belochub fixed.

nechaido:
Landed in 7441fd1d6e93a2227a08abfdfb36b1f755472e77.

Issue #154, opened 2017-05-10
Build native plugins using multiple threads


nechaido:
Currently native plugins are being build using only one thread.
It would be great if we used all accessible threads if it is possible.

aqrln:
@belochub feel free to take this if you are willing to, otherwise I'll look into it tomorrow or maybe today.

PR #185, opened 2017-05-26
build: run native addon build in parallel

belochub:
Fixes: https://github.com/metarhia/jstp/issues/154

### DIFF
--- a/tools/build-native.js
+++ b/tools/build-native.js
@@ -18,7 +18,7 @@ fs.access('build', (error) => {
   }
 
   const nodeGyp = childProcess.spawn('node-gyp', [action, jobs],
-    { shell: true });
+                                     { shell: true });
   const errorLines = [];
 
   nodeGyp.on('error', () => {
--- a/tools/build-native.js
+++ b/tools/build-native.js
@@ -3,19 +3,22 @@
 'use strict';
 
 const fs = require('fs');
+const os = require('os');
 const childProcess = require('child_process');
 
 const EXIT_SUCCESS = 0;
 const EXIT_FAIL = 1;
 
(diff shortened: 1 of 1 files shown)


tools/build-native.js
```diff
@@ -3,19 +3,22 @@
...
+const jobs = '-j' + os.cpus().length;
 
 fs.access('build', (error) => {
   if (error) {
     action = 'rebuild';
   }
 
-  const nodeGyp = childProcess.spawn('node-gyp', [action], { shell: true });
+  const nodeGyp = childProcess.spawn('node-gyp', [action, jobs],
+    { shell: true });
```
nechaido: I'd prefer each argument to be on new line, like this:
```js
const nodeGyp = childProcess.spawn(
  'node-gyp',
  [action, jobs],
  {shell: true }
);
```

belochub:
@lundibundi, I've added a possibility to use JOBS environment variable, PTAL.
About the second point: I don't think that there is any need to limit parallel jobs, most people even use job counts exceeding processor count.

tools/build-native.js
aqrln: Either that ↑ or move the last argument all the way right to the column where the first one starts.

aqrln: For the clarity: I'd go for the latter option but I'm fine with it either way.

lundibundi:
@belochub thanks, I wrote it as I personally tend to limit j flag to smth like that as I prefer to have at least one or 2 cpus free on my machine to allow seamless workflow even during system update/build/npm install process.

aqrln:
Landed in 367fcd5e749b5e8b1791cc4c1a825ac5799352cc.

PR #186, opened 2017-05-27
deps: update deps and add package-lock.json

aqrln:
A new round of dependencies updates!

The only semver-major update is `babel-loader@7` with dropped support for old Node.js and webpack versions as the only breaking changes. The other available major update is `chai@4`, but I had to leave `chai@3.5` because `chai-spies` package is still incompatible with it, and that doesn't matter anyway since we are migrating to `tap`.

Aside from that, npm 5 was released yesterday 🎉, so add `package-lock.json`.


aqrln:
The `package-lock.json` in this PR represents a broken dependency tree. Blocking this PR until the issues with [`github-slugger`](https://www.npmjs.com/package/github-slugger) are resolved in npm 5.

aqrln:
The issue with `github-slugger` is now resolved, removing the `blocked` label.

aqrln:
@metarhia/jstp-core ping

aqrln:
Landed in 9e1952c9f4175cafc00cb733c2bada63dd6881fc.

PR #187, opened 2017-05-28
lib: change event signature

lundibundi:
Changes event message signature to always send an array
of arbitrary arguments. This makes it work same way as
EventEmitter does and also makes its usage similar to
call action.

Also the tests are just plainly awesome, but we will rewrite them soon anyway.

### DIFF
--- a/lib/remote-proxy.js
+++ b/lib/remote-proxy.js
@@ -37,8 +37,7 @@ RemoteProxy.prototype.emit = function(eventName, ...eventArgs) {
 //   eventName - name of an event
 //   eventArgs - array of event arguments
 //
-RemoteProxy.prototype._emitLocal = function(eventName, eventArgs) {
-  if (!eventArgs) eventArgs = [];
+RemoteProxy.prototype._emitLocal = function(eventName, eventArgs = []) {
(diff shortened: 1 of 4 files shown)


lib/remote-proxy.js
```diff
@@ -24,22 +24,22 @@ function RemoteProxy(connection, interfaceName, methods = []) {
...
+  this._emitLocal(eventName, eventArgs);
+};
 
-  events.EventEmitter.prototype.emit.call(this, eventName, eventArgs);
+// Emit local event.
+//   eventName - name of an event
+//   eventArgs - array of event arguments
+//
+RemoteProxy.prototype._emitLocal = function(eventName, eventArgs) {
+  if (!eventArgs) eventArgs = [];
```
belochub: You should use default parameters instead of this.

aqrln:
@belochub as far as I can tell from the in-person conversation, you weren't extremely happy with this change. Did you change your mind or...?

Defensively marking this as blocked until I'm sure there is consensus about this.

belochub:
@aqrln, yes, it is okay, I approved the PR after @lundibundi explained to me the point of it.

lundibundi:
Landed in 852b46c0ac488eb36a8733f18bc6a2b1fed1f153.

PR #189, opened 2017-05-29
tcp: fix listening on port 0

aqrln:
Fix invalid condition that conflated port 0 and the absence of port.


aqrln:
IMO, this change is a trivial enough to land immediately after the CI and a sign-off or two.

aqrln:
Landed in 804f676fc7e429c4446ac88d5059cbd2bd04514e.

PR #190, opened 2017-05-29
lib: add address method to Server

lundibundi:
This allows to properly get server address parameters
without the need to break abstraction to get them from
low level server.

### DIFF
--- a/test/node/get-server-address-ipc.js
+++ b/test/node/get-server-address-ipc.js
@@ -1,0 +1,39 @@
+'use strict';
+
+const test = require('tap');
+const fs = require('fs');
+
+const jstp = require('../..');
+
+const app = new jstp.Application('name', {});
+
+const path = process.platform === 'win32' ?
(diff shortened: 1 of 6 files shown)


belochub:
@lundibundi, you shouldn't add `test` label when you include tests connected to implemented functionality.

lundibundi:
@belochub sure.

aqrln:
Landed in 21ff8b7347d7de6bb3d06eb4e0c2e68ad2edc0d2.

test/node/get-server-address-ipc.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
```
aqrln: @belochub btw, after some thinking about it I'm now on your side about the fate of these tests if/when we'll be returning subclasses of Node.js core servers. Not that I'll be determined to remove them and I still think that extra tests aren't a bad thing even if there's not much need in them, but both keeping and removing these tests will sound fine to me.

Issue #155, opened 2017-05-10
Fix Karma using hardcoded browsers


nechaido:
Currently build fails on my PC due to this issue:
```
10 05 2017 19:29:18.230:ERROR [launcher]: No binary for ChromeCanary browser on your platform.
  Please, set "CHROME_CANARY_BIN" env variable
```
<details><summary>npm output</summary><p>

```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! metarhia-jstp@0.7.0-pre test-unit-browser: `karma start`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the metarhia-jstp@0.7.0-pre test-unit-browser script 'karma start'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the metarhia-jstp package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     karma start
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs metarhia-jstp
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls metarhia-jstp
npm ERR! There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/nechaido/.npm/_logs/2017-05-10T16_29_23_391Z-debug.log
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! metarhia-jstp@0.7.0-pre test-unit: `npm run test-unit-node && npm run test-unit-browser`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the metarhia-jstp@0.7.0-pre test-unit script 'npm run test-unit-node && npm run test-unit-browser'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the metarhia-jstp package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     npm run test-unit-node && npm run test-unit-browser
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs metarhia-jstp
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls metarhia-jstp
npm ERR! There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
[...]

Issue #149, opened 2017-05-08
Fix Karma tests on Windows


aqrln:
As far as I know (at least, last time I tried) our Karma config isn't quite Windows-friendly. Not only it doesn't run the tests in IE and Edge, but it also fails (failed?) to start at all.

PR #193, opened 2017-05-30
test: autodetect available browsers in karma

lundibundi:
- [x] use karma-detect-browsers to autodetect available browsers
- [x] add edge and opera browsers to karma configuration

I haven't added [karma-ie-launcher](https://github.com/karma-runner/karma-ie-launcher)  as it has been last updated quite a while ago and has a lot of issues, so I'm not sure if we should add it.

Fixes: https://github.com/metarhia/jstp/issues/155


.travis.yml
```diff
@@ -32,6 +32,7 @@ addons:
       - g++-6
   firefox: latest
 before_script:
+  - export CHROME_BIN=chromium-browser
```
aqrln: I don't thing it will work because of sandboxing issues, but let's see what Travis says. Maybe it will work this time.

package.json
```diff
@@ -46,11 +46,14 @@
     "eslint": "^3.12.2",
     "eslint-plugin-import": "^2.2.0",
     "karma": "^1.3.0",
+    "karma-mocha": "^1.3.0",
```
aqrln: The order of packages is messed up here. Did you edit it manually?

.travis.yml
lundibundi: @aqrln yeah, I wasn't sure either but it was worth a try. It failed. Any ideas?

package.json
lundibundi: @aqrln yeah, completely forgot about npm commands =). Got used to editing config files manually that much

aqrln:
I suppose it should also fix https://github.com/metarhia/jstp/issues/149.

lundibundi:
@aqrln I think so too but I haven't tested it therefore I didn't add it. I'll probably test it later before landing it and add fixes if it works.

aqrln:
@lundibundi Windows fails with `Cannot load browser "IE": it is not registered`.
`npm i -D karma-ie-launcher` fixes this. The tests themselves fail with `SyntaxError`s on IE and Edge though, but that's a matter of a different PR.

aqrln:
Oh, and can you please rebase onto master and update `package-lock.json`?

lundibundi:
@aqrln thanks for testing it on windows. No problem.

lundibundi:
@aqrln I've updated package.json and added fixes 149, as In my opinion the issue was about starting karma tests on windows and we have to create separate issue dedicated to fixing the errors detected.

Also Arch Linux still hasn't updated npm version, could you please land it and update `package-lock.json` upon landing?

aqrln:
@lundibundi 

> Also Arch Linux still hasn't updated npm version, could you please land it and update package-lock.json upon landing?

`npm i -g npm@5` ;)
But np, I can update it myself.

lundibundi:
@aqrln I obviously know of that, but it'll probably break pacman packet and I don't really want to fix with that. :smile:

aqrln:
@lundibundi that won't break the pacman package. npm bundled with Node.js and npm installed as a regular global package reside in different directories.

aqrln:
Landed in 6339bb0ce01847fd90f53d94e4581e0d78a2226b.

PR #198, opened 2017-05-31
test: migrate inspect and remote proxy tests to tap

nechaido:
Refs: #145
Refs: #146 
Refs: #53.


test/node/connection-inspect.js
```diff
@@ -0,0 +1,57 @@
...
+  );
+});
+
+test.test('must throw error if interface does not exist', (test) => {
+  client.connectAndInspect(app.name, null, null, ['__nonexistent_-interface'],
+    (error) => {
+      test.assert(error, 'must return an error');
+      test.equal(error.code, jstp.ERR_INTERFACE_NOT_FOUND,
+        'error must be an ERR_INTERFACE_NOT_FOUND'
+      );
```
aqrln: Maybe merge this line with the previous one?

nechaido:
Blocked until #190.

aqrln:
Still LGTM.

aqrln:
Oh, only one more thing: can please test `connection.inspectInterface()` directly too?

test/fixtures/application.js
```diff
@@ -0,0 +1,26 @@
...
+    sayHi(connection, name, callback) {
+      callback(null, `Hi, ${name}!`);
+    },
+    say(connection, word, callback) {
+      callback(null, word);
+    }
+  },
+  someOtherService: {
+    method(connection, argument, callback) {
+      callback(argument);
```
belochub: Is returning argument as an error intentional here?

test/node/connection-inspect.js
```diff
@@ -0,0 +1,97 @@
...
+  client.disconnect();
+  server.close();
+  done();
+});
+
+test.test('must process inspect packets', (test) => {
+  const expectedInterfaces = Object.keys(app.interfaces);
+  let expectedTests = 1;
+  expectedInterfaces.forEach((iface) => {
+    expectedTests += 1;
```
belochub: May you merge this with the next line? Like this:
```js
expectedTests += Object.keys(app.interfaces[iface]).length + 1;
```

test/node/connection-inspect.js
```diff
@@ -0,0 +1,97 @@
...
+
+  test.plan(expectedTests);
+  client.connectAndHandshake(app.name, null, null,
+    (error, connection) => {
+      test.assertNot(error, 'must connect to server');
+
+      expectedInterfaces.forEach((iface) => {
+        connection.inspectInterface(iface, (error, methods) => {
+          test.assertNot(error, `must inspect ${iface}`);
+          Object.keys(app.interfaces[iface]).forEach((method) => {
```
belochub: You may avoid using braces and parenthesis here, there is just one statement in this function.

test/node/connection-inspect.js
```diff
@@ -0,0 +1,97 @@
...
+test.test('must generate remote api after inspect', (test) => {
+  const expectedInterfaces = Object.keys(app.interfaces);
+
+  client.connectAndInspect(app.name, null, null, expectedInterfaces,
+    (error, connection, api) => {
+      test.assertNot(error, 'must connect to server');
+
+      expectedInterfaces.forEach((iface) => {
+        test.assert(iface in api, `api must include '${iface}'`);
+        Object.keys(app.interfaces[iface]).forEach((method) => {
```
belochub: Ditto.

test/fixtures/application.js
nechaido: It's stub that is never called, but I'll fix it to behave properly if called

test/node/connection-inspect.js
nechaido: I'd like to leave it this way because it "separates" check count for interfaces and methods.

belochub: @nechaido, if that's the case, may you add a comment explaining this, to avoid confusion in the future?
Also, why not just add `expectedInterfaces.length` to `expectedTests` immediately after initialization, or even initialize `expectedTests` with  `expectedInterfaces.length + 1`?

nechaido: @belochub I though It'll be more obvious, seems It's not, also I'm strongly opposed to leaving a comment. It should be written in a more obvious way that needs no explanation. Maybe I'll stop on my initial idea: also proposed by you:
```
let expectedTests = 1 + expectedInterfaces.length;
expectedTests +=  expectedInterfaces.reduce((methodsCount, iface) => {
    methodsCount += Object.keys(app.interfaces[iface]).length;
  });
```
Maybe I should merge this into one expression.
@metarhia/jstp do you have any ideas concerning this issue?

nechaido: Or even:
```
let expectedTests = expectedInterfaces.reduce((tests, iface) => {
    tests += 1 + Object.keys(app.interfaces[iface]).length;
}, 1);
```

lundibundi: The latter looks better to me, but I think it's better to add a comment explaining this.

nechaido:
@aqrln added.

test/node/connection-inspect.js
```diff
@@ -0,0 +1,97 @@
...
+        test.assertNot(error, 'remote method must be called without error');
+        test.strictEqual(result, word,
+          'remote method should return a proper result');
+        test.end();
+      });
+    }
+  );
+});
+
+test.test('must throw error if interface does not exist', (test) => {
```
aqrln: It doesn't really throw, it just returns an error.

nechaido:
Landed in 6be836e6fbe3944ead8dbb5e6a0aa8a13d183bd3.

Issue #175, opened 2017-05-23
Make number parsing in native parser implementation JSON5 compliant


belochub:
- [x] Fix number parsing bugs (https://github.com/metarhia/jstp/pull/207, https://github.com/metarhia/jstp/pull/208, https://github.com/metarhia/jstp/pull/221)
- [x] Implement parsing of `NaN` and `Infinity` (https://github.com/metarhia/jstp/pull/201)

Refs: https://github.com/metarhia/jstp/issues/152

belochub:
Closed via https://github.com/metarhia/jstp/pull/201, https://github.com/metarhia/jstp/pull/207, https://github.com/metarhia/jstp/pull/208, https://github.com/metarhia/jstp/pull/221.

Issue #69, opened 2017-02-10
Make parser errors more detailed and correct for some of the cases


belochub:
At the moment most of the errors thrown by parser are not informational and sometimes even wrong.

aqrln:
@belochub are you still up to working on this?

PR #201, opened 2017-06-01
parser: implement NaN and Infinity parsing

belochub:
Also fix not throwing on some of the incorrect number inputs.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -4,6 +4,7 @@
 #include "jsrs_parser.h"
 
 #include <cctype>
+#include <cmath>
 #include <cstddef>
 #include <cstdlib>
 #include <cstring>
@@ -18,6 +19,8 @@ using std::function;
 using std::isalnum;
 using std::isalpha;
 using std::isdigit;
+using std::isinf;
+using std::isnan;
 using std::isxdigit;
 using std::memset;
 using std::ptrdiff_t;
@@ -148,6 +151,11 @@ static bool GetType(const char* begin, const char* end, Type* type) {
       if (begin + 9 <= end) {
         result = (strncmp(begin, "undefined", 9) == 0);
       }
+      break;
+    }
+    case 'N':
+    case 'I': {
+      *type = Type::kNumber;
       break;
     }
     default: {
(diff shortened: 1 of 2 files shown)


aqrln:
@lundibundi +1

src/jsrs_parser.cc
```diff
@@ -150,6 +151,11 @@ static bool GetType(const char* begin, const char* end, Type* type) {
       }
       break;
     }
+    case 'N':
+    case 'I': {
+      *type = Type::kNumber;
```
aqrln: Not sure about this for the reasons @lundibundi has mentioned...

belochub: @aqrln, why'd you put an ellipsis in the end of your comment?
Also, the problem stated by @lundibundi has nothing to do with this line, it is just incorrect usage of the `strtol` function, and I will fix this.

belochub: Oops, I am sorry, I was really talking about `atof` in the previous comment, not `strtol`.
`strtod` has to be used instead of it.

src/jsrs_parser.cc
```diff
@@ -318,32 +326,46 @@ MaybeLocal<Value> ParseNumber(Isolate*    isolate,
...
+  if (negate_result) {
+    number = -number;
   }
-  *size = i;
-  return result;
+
+  // strictly allow only "NaN" and "Infinity"
+  if (isnan(number)) {
+    if (strncmp(begin + 1, "aN", 2) != 0) {
+      THROW_EXCEPTION(SyntaxError, "Invalid format: expected NaN");
```
lundibundi: I think it's better to say 'Invalid number format: ...'.

src/jsrs_parser.cc
```diff
@@ -318,32 +326,46 @@ MaybeLocal<Value> ParseNumber(Isolate*    isolate,
...
+
+  if (negate_result) {
+    number = -number;
   }
-  *size = i;
-  return result;
+
+  // strictly allow only "NaN" and "Infinity"
+  if (isnan(number)) {
+    if (strncmp(begin + 1, "aN", 2) != 0) {
```
lundibundi: wow, such a pain =(

src/jsrs_parser.cc
```diff
@@ -318,32 +326,46 @@ MaybeLocal<Value> ParseNumber(Isolate*    isolate,
...
+      THROW_EXCEPTION(SyntaxError, "Invalid format: expected NaN");
+      return MaybeLocal<Value>();
+    }
+  } else if (isinf(number)) {
+    if (strncmp(begin + 1, "nfinity", 7) != 0) {
+      THROW_EXCEPTION(SyntaxError, "Invalid format: expected Infinity");
```
lundibundi: ditto.

src/jsrs_parser.cc
belochub: I wrote it like that to be consistent with the other error messages, it still isn't perfect everywhere, but hopefully I will fix it someday (see #69).

belochub: What I wanted to say is that it is an issue for another PR which will improve error reporting in the native parser.

src/jsrs_parser.cc
belochub: What's so painful here? I just avoided testing for the same character twice here, that's it.

lundibundi: The need to actually check this.

src/jsrs_parser.cc
lundibundi: @belochub Ok, let's hope it will be done eventually =)

belochub:
Landed in 8e87a284a502398b6a7459a3a71e6a5b76df049e.

PR #208, opened 2017-06-03
parser: fix parsing of big integer values

belochub:
Implement additional function for parsing integer values which
cannot be parsed using standard library functions.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -4,8 +4,10 @@
 #include "jsrs_parser.h"
 
 #include <cctype>
+#include <cerrno>
 #include <cmath>
 #include <cstddef>
+#include <cstdint>
 #include <cstdlib>
 #include <cstring>
 #include <functional>
@@ -28,6 +30,7 @@ using std::size_t;
 using std::strncmp;
 using std::strncpy;
 using std::strtol;
+using std::toupper;
 
 using v8::Array;
 using v8::False;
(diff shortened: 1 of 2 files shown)


belochub:
@aqrln, ping.

src/jsrs_parser.cc
```diff
@@ -375,12 +378,48 @@ Local<Value> ParseIntegerNumber(Isolate*    isolate,
...
+  char base_alpha_count = base > 10 ? base - 10 : 0;
+  for (size_t i = 0; i < *size; i++) {
+    current_digit = toupper(begin[i]);
+    if ((current_digit < '0' || current_digit >= '0' + base_digit_count) &&
+        (current_digit < 'A' || current_digit >= 'A' + base_alpha_count)) {
+      *size = i;
+      break;
+    }
+    current_digit_value = current_digit <= '9' ? current_digit - '0' :
+      current_digit - 'A' + 10;
```
aqrln: I believe this should be indented with four spaces.

belochub: @aqrln, `cpplint` didn't complain about it, so I left it like that.
What do you say about moving it to the right, right after the question sign?

aqrln: @belochub sounds good to me

belochub:
Landed in https://github.com/metarhia/jstp/commit/417ef60bdde3e9a68349399f6e7ef2b1e7bb4f1a.

Issue #192, opened 2017-05-29
Use inheritance instead of composition in server adapters


belochub:
At the moment all of our Server adapters (which are passed to its constructor as `rawServer` argument) are using composition to store the original server object, which is being created in the same place using methods similar to those original server object has.
It results in a lot of boilerplate code and usage of workarounds which can be avoided by inheriting the adapters from the classes used in an underlying implementation now.

This will obviate PRs like https://github.com/metarhia/jstp/pull/190 and should we decide to go with the proposed approach, it will make sense to close the PR.

@metarhia/jstp-core, what is your opinion on the topic?

PR #209, opened 2017-06-03
lib: refactor server and client api

lundibundi:
As of now I mostly finished with the refactorings so feel free to comment on the changes in API. The next step is to change all available tests (those written in tap probably) and make sure this implementation passes them.

Fixes: https://github.com/metarhia/jstp/issues/192
Fixes: https://github.com/metarhia/jstp/issues/146


lib/socket.js
```diff
@@ -0,0 +1,119 @@
...
+      this.emit('packet', packets[i]);
+    }
+
+    if (this._buffer.length > MAX_PACKET_SIZE) {
+      this.emit('error', new Error('Maximal packet size exceeded'));
+    }
+  }
+}
+
+const socketFactory = connect => (...options) => {
```
nechaido: It is not actually a socket factory which is confusing. We should try to find more appropriate name.
@metarhia/jstp-core what do you think?

lib/socket.js
```diff
@@ -0,0 +1,119 @@
...
+};
+
+// Create a function that will be bound to socketFactory that will
+// produce JSTP connection bound to a socket created with socketFactory
+//   connect - function that will be called with ...options
+//             and must return object that emits events 'connect' and 'error'
+//
+// see transportCommon.createConnectionFactory
+//
+const createConnectionFactory = connect =>
```
nechaido: This name is a bit misleading too, IMHO.

lib/socket.js
```diff
@@ -0,0 +1,119 @@
...
+    socketFactory(connect),
+    socket => new Transport(socket)
+  );
+
+// Same as createConnectionFactory but will also perform inspect of specified
+// interfaces.
+//
+// see transportCommon.createConnectionAndInspectFactory
+//
+const createConnectionAndInspectFactory = connect =>
```
nechaido: Ditto.

test/node/connection-call.js
```diff
@@ -7,70 +7,68 @@ const jstp = require('../..');
...
-    });
+    test.end();
   });
 });
 
 test.test('must return error on call to nonexistent interface', (test) => {
-  client.connectAndHandshake(app.name, null, null, (error, connection) => {
-    connection.callMethod(
+  connection.callMethod(
       '__nonexistent_interface__',
```
nechaido: The indentation is off here.

test/node/connection-call.js
```diff
@@ -80,19 +78,14 @@ test.test('must return error on call to nonexistent interface', (test) => {
...
 test.test('must return error on call to nonexistent method', (test) => {
-  client.connectAndHandshake(app.name, null, null, (error, connection) => {
-    connection.callMethod('calculator', '__nonexistent_method__', [],
-      (error) => {
-        test.assert(error, 'callMethod must return an error');
-        test.equal(error.code, jstp.ERR_METHOD_NOT_FOUND,
+  connection.callMethod('calculator', '__nonexistent_method__', [], (error) => {
+    test.assert(error, 'callMethod must return an error');
+    test.equal(error.code, jstp.ERR_METHOD_NOT_FOUND,
           'error must be an ERR_METHOD_NOT_FOUND');
```
nechaido: Ditto.

test/node/connection-inspect.js
```diff
@@ -31,28 +30,30 @@ test.test('must process inspect packets', (test) => {
...
 
-      expectedInterfaces.forEach((iface) => {
-        connection.inspectInterface(iface, (error, methods) => {
-          test.assertNot(error, `must inspect ${iface}`);
-          Object.keys(app.interfaces[iface]).forEach(method =>
+    expectedInterfaces.forEach((iface) => {
+      connection.inspectInterface(iface, (error, methods) => {
+        test.assertNot(error, `must inspect ${iface}`);
+        Object.keys(app.interfaces[iface]).forEach(method =>
             test.assert(method in methods,
```
nechaido: Ditto.

lib/common.js
```diff
@@ -38,6 +38,27 @@ common.safeRequire = (moduleName) => {
   }
 };
 
+// Mixin source methods to target without overriding existing methods
+// for ES6 classes.
+//
+common.mixin = (target, source) => {
+  Object.getOwnPropertyNames(source.prototype).forEach((method) => {
```
lundibundi: @nechaido fine with me.

lib/net.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+const net = require('net');
+
+const sock = require('./socket');
```
lundibundi: @nechaido okay, its legacy from prev implementation and I didn't really care.

lib/socket.js
lundibundi: I agree with @nechaido, but couldn't come up with a better name.

test/node/connection-call.js
lundibundi: @nechaido thanks.

test/node/connection-call.js
lundibundi: @nechaido thanks.

test/node/connection-inspect.js
lundibundi: @nechaido thanks, missed them.

lib/common.js
```diff
@@ -38,6 +38,27 @@ common.safeRequire = (moduleName) => {
...
+common.mixin = (target, source) => {
+  Object.getOwnPropertyNames(source.prototype).forEach((property) => {
+    if (!target.prototype[property]) {
+      target.prototype[property] = source.prototype[property];
+    }
+  });
+};
+
+// If last element of the array args is a function then
+// pops the array and returns that function else return null
```
aqrln: Nit: wrong tense in the last "return" (it should be "returns").

lib/common.js
```diff
@@ -38,6 +38,27 @@ common.safeRequire = (moduleName) => {
...
+//
+common.extractCallback = (args) => {
+  if (args.length > 0 && typeof args[args.length - 1] === 'function') {
+    return args.pop();
```
aqrln: This will modify arrays that a user passes to the functions of jstp's public API, right?

lib/server.js
```diff
@@ -1,131 +1,99 @@
...
-  this.rawServer = rawServer;
   this.applications = applications;
-  this.startSession = authCallback || simpleAuth.startSession;
   this.heartbeatInterval = heartbeatInterval;
 
+  if (typeof authPolicy === 'function') {
+    this.startSession = authPolicy;
+  } else {
+    if (!authPolicy) authPolicy = new SimpleAuthPolicy();
+    this.startSession = (...args) => authPolicy.startSession(...args);
```
aqrln: Using `authPolicy.startSession.bind(authPolicy)` might be better in this case.

lib/simple-auth-policy.js
```diff
@@ -0,0 +1,54 @@
...
+    const sessionId = generateUuid4();
+    this.emit('session', sessionId, connection, application);
+
+    callback(null, null, sessionId);
+  }
+};
+
+// Generate UUID v4
+//
+function generateUuid4() {
```
aqrln: This function shouldn't be in this package, it falls outside its scope. While you are here and doing some refactoring, maybe we should drop in this PR? Exactly this function (which I replaced the one from Impress with at some point) is now in `metarhia-common`, I believe; and there are other npm packages too, like `uuid`.

lib/simple-connect-policy.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const EventEmitter = require('events').EventEmitter;
+
+// Simple generic connection provider. Used for user-side connection.
+// You are free to implement whatever suits your needs instead.
+// Sends handshake with login/password if provided otherwise sends
+// anonymous handshake.
+//
+module.exports = class SimpleConnectPolicy extends EventEmitter {
```
aqrln: Does it really even need to be a class, not to mention an EventEmitter (I don't see any EventEmitter functionality used here)? Requiring this and the auth policy to be just functions seems both simpler and more flexible to me (one can always pass a bound method if they need it to be stateful).

lib/common.js
```diff
@@ -38,6 +38,27 @@ common.safeRequire = (moduleName) => {
   }
 };
 
+// Mixin source methods to target without overriding existing methods
+// for ES6 classes.
+//
+common.mixin = (target, source) => {
```
aqrln: This function needs tests. They can be added in a follow-up PR, though.

lib/common.js
```diff
@@ -38,6 +38,27 @@ common.safeRequire = (moduleName) => {
...
+      target.prototype[method] = source.prototype[method];
+    }
+  });
+};
+
+// If last element of the array args is a function then
+// pops the array and returns that function else return null
+//
+common.extractCallback = (args) => {
+  if (args.length > 0 && typeof args[args.length - 1] === 'function') {
```
aqrln: The first check is superfluous.

jstp.js
```diff
@@ -13,9 +13,11 @@ jstp.RemoteProxy = require('./lib/remote-proxy');
...
 jstp.Server = require('./lib/server');
 
-jstp.tcp = require('./lib/transport.tcp');
-jstp.ipc = require('./lib/transport.ipc');
-jstp.ws = require('./lib/transport.ws');
+jstp.net = require('./lib/net');
+jstp.tls = require('./lib/tls');
+jstp.ws = require('./lib/ws');
+jstp.wss = require('./lib/wss');
+jstp.wsBrowser = require('./lib/ws-browser');
```
aqrln: We should require just `ws` here, resolving it to the actual implementation is up to the module loader (either Node's or webpack's).

lib/net.js
aqrln: Having `socket` as the name of the parameter of `createTransport()` is more important.

lib/net.js
```diff
@@ -0,0 +1,61 @@
...
+  //   options - an object that contains applications array or
+  //             index and can optionally contain authPolicy and
+  //             heartbeatInterval (see jstp.Server).
+  //             Also it will be passed directly to node net.Server.
+  //   listener - request listener that will be passed directly to
+  //              net.Server
+  //
+  // See node net.Server for options docs.
+  //
+  constructor(options, listener) {
```
aqrln: This shouldn't be a listener of a raw TCP connection, that's practically useless and is a leaky abstraction. The listener function that is passed to `super()` must be an internal function, while the optional `listener` argument of this constructor should be attached as a listener of the `'connect'` event, which shall be introduced.

lib/transport-common.js
```diff
@@ -0,0 +1,105 @@
...
+      });
+    });
+  };
+};
+
+// Utility method to create function to produce servers with
+// serverFactory.
+// If options is an array then wraps it as { applications: options }.
+//
+const createServerFactory = serverFactory => (options, ...other) => {
```
aqrln: This function should take a class to avoid creating unnecessary closures in concrete server implementations.

lib/socket.js
```diff
@@ -0,0 +1,119 @@
...
+      this.emit('error', new Error('Maximal packet size exceeded'));
+    }
+  }
+}
+
+const socketFactory = connect => (...options) => {
+  const callback = common.extractCallback(options);
+  const socket = connect(...options);
+  socket.once('connect', () => callback(null, socket));
+  socket.once('error', error => callback(error));
```
aqrln: You need to check that `callback` has never been called here.

lib/transport-common.js
```diff
@@ -0,0 +1,105 @@
...
+const createConnectionFactory = (
+  connFactory, transportFactory
+) => (appName, client, ...options) => {
+  const callback = common.extractCallback(options);
+  connFactory(...options, (error, rawConnection) => {
+    if (error) return callback(error);
+
+    const transport = transportFactory(rawConnection);
+    if (!client) {
+      client = { application: new Application('jstp', {}) };
```
aqrln: Can you please add `, connectPolicy: null` here so that the object shape does not change at line 44?

aqrln:
@lundibundi can you please rebase and resolve the conflicts?

lib/connection.js
```diff
@@ -1,16 +1,13 @@
 'use strict';
 
-const events = require('events');
+const EventEmitter = require('events').EventEmitter;
```
aqrln: Tiny nit: `const { EventEmitter } = require('events');`

lib/connection.js
```diff
@@ -27,607 +24,606 @@ let PACKET_HANDLERS = null;
...
-  this._callbacks[packetId] = (error, ...methods) => {
-    if (error) {
+  // Send a handshake packet over the connection
+  //   appName - name of an application to connect to
+  //   login - user name (optional)
+  //   password - user password (optional)
+  //   callback - callback function to invoke after the handshake is completed
+  //
+  handshake(appName, login, password, callback) {
+    const packet = login && password ? this.createPacket('handshake', appName,
```
aqrln: Can you please break the lines after `?` and `:` characters instead of between arguments?

lib/connection.js
```diff
@@ -27,607 +24,606 @@ let PACKET_HANDLERS = null;
...
-Connection.prototype._onTimeout = function() {
-  this.emit('_timeout');
-};
+  // Create a JSTP packet
+  //   kind - packet kind
+  //   target - name of an interface or an application (optional)
+  //   verb - action specific for different packet kinds
+  //   args - action arguments
+  //
+  createPacket(kind, target, verb, args) {
```
aqrln: Is there any reason this method is public? If not, can you please prepend an underscore to its name?

lib/socket.js
aqrln: Maybe something like `newCreateConnectionFn`?

lib/simple-connect-policy.js
lundibundi: @aqrln yeah, EventEmitter is redundant there, thanks. Auth policy can be just function and I'll allow Connect policy to also be a function as that is indeed more flexible.

lib/applications.js
```diff
@@ -10,47 +10,46 @@ module.exports = apps;
...
-  return appInterface && Object.keys(appInterface);
-};
+  // Get an array of methods of an interface
+  //   interfaceName - name of the interface to inspect
+  //
+  getMethods(interfaceName) {
+    const appInterface = this.api[interfaceName];
+    return appInterface && Object.keys(appInterface);
+  }
+}
```
lundibundi: Sure.

lib/connection.js
lundibundi: Ok.

lib/connection.js
lundibundi: Sure, it looks much better.

lib/connection.js
lundibundi: No reason, I just didn't change it from the way it was. Will make it private.

lib/socket.js
lundibundi: Ok.

lundibundi:
Closed via 34cd6ee133196475154651ef9fcab9988828463d and a09dfe86dff1742015c687f90d9af4626cd488a9.

PR #210, opened 2017-06-05
cli: fix exception on unestablished connection

lundibundi:
When you made connect but connection was not established yet cli would fail on call and event commands with exception.
Not sure about message in new error so feel free to suggest another one.

### DIFF
--- a/lib/cli/command-processor.js
+++ b/lib/cli/command-processor.js
@@ -88,12 +88,18 @@ module.exports = class CommandProcessor extends EventEmitter {
   }
 
   call(interfaceName, methodName, args, callback) {
-    if (!this.cli.client) return callback(new Error('Not connected'));
+    if (!this.cli.connection) {
+      if (this.cli.client) return callback(new Error('Connection in progress'));
+      return callback(new Error('Not connected'));
+    }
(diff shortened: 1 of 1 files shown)


lib/cli/command-processor.js
```diff
@@ -88,12 +88,18 @@ module.exports = class CommandProcessor extends EventEmitter {
   }
 
   call(interfaceName, methodName, args, callback) {
-    if (!this.cli.client) return callback(new Error('Not connected'));
```
belochub: @lundibundi
What do you say about removing redundant indentation by leaving line 91 as it was, and just adding this:
```js
if (!this.cli.connection) return callback(new Error('Connection in progress'));
```

lundibundi: @belochub it doesn't fit, does it?

belochub: @lundibundi, it doesn't fit where?

lundibundi: @belochub done.

belochub:
Landed in https://github.com/metarhia/jstp/commit/6cb62a0251f90cb557973f22da1b35ea0c7819f4.

PR #213, opened 2017-06-07
test: remove mocha and karma

nechaido:
Removes obsolete mocha tests and karma because it only run mocha tests. Karma will returned into project after adapting current tests to in browser.


test/node/connection-handshake.js
```diff
@@ -7,7 +7,7 @@ const jstp = require('../..');
 const app = require('../fixtures/application');
 const application = new jstp.Application(app.name, app.interfaces);
 
-const Transport = require('../unit/mock/transport');
+const Transport = require('../util/transport.mock');
```
aqrln: I'd prefer it to be `'../mock/transport'`.

package.json
```diff
@@ -41,34 +41,18 @@
     "babel-loader": "^7.0.0",
     "babel-polyfill": "^6.23.0",
     "babel-preset-es2015": "^6.24.1",
-    "chai": "^3.5.0",
```
aqrln: Can you please update the `package-lock.json`? It can as well be done by whoever lands this though.

test/node/connection-handshake.js
```diff
@@ -7,7 +7,7 @@ const jstp = require('../..');
 const app = require('../fixtures/application');
 const application = new jstp.Application(app.name, app.interfaces);
 
-const Transport = require('../util/transport.mock');
+const Transport = require('../mock/transport.mock');
```
aqrln: I don't think we need to duplicate the word "mock". `../mock/transport` will work just fine.

PR #214, opened 2017-06-07
dist: add coverage task

nechaido:


package.json
```diff
@@ -56,6 +56,7 @@
     "karma-safari-launcher": "^1.0.0",
     "karma-webpack": "^2.0.3",
     "mocha": "^3.4.2",
+    "nyc": "^11.0.2",
```
aqrln: `package-lock.json` has to be updated.

belochub: @aqrln, it already includes this package.

aqrln: @belochub oh, well then.

aqrln: @belochub nope, I checked it out. It only includes `nyc@10.3.2` while this PR requires `nyc@11`.

belochub: @aqrln, but after looking at it one more time, the major version of this package in `package-lock.json` is 10, not 11, so it should still probably be updated accordingly.

aqrln: OTOH the newer version of `tap` relies upon `nyc@11` too, so we might just update the existing dependencies.

nechaido:
@aqrln @belochub I've added commit that updates `tap` and adds `nyc` to `package-lock.json`

nechaido:
Landed in 015f5d9fc06652806dccffd323cd9f8fac67b315 and 3fc41dab4fa4fc96d7ee1a739727f330166042a9.

Issue #188, opened 2017-05-29
Remove unnecessary output from common.safeRequire()


nechaido:
Currently we have this when we run tests:
```
39 passing (51ms)

test/node/common-do-nothing.js ........................ 1/1 333ms
test/node/common-forward-event.js ..................... 4/4
Error: Cannot find module '__non_existing_module__'
test/node/common-safe-require.js ...................... 2/2
test/node/get-server-address-ipc.js ................... 2/2
test/node/get-server-address-tcp.js ................... 2/2
test/node/get-server-address-ws.js .................... 2/2
test/node/message-parser.js ........................... 8/8
test/node/remote-error-constructor.js ............... 20/20
test/node/remote-error-default-messages.js ............ 1/1 359ms
test/node/remote-error-from-jstp-array.js ........... 20/20
test/node/remote-error-get-jstp-array.js .............. 6/6
test/node/remote-error-to-jstp-array.js ............... 4/4
test/node/serde.js .................................. 65/65
total ............................................. 137/137
```
I think the line: `Error: Cannot find module '__non_existing_module__'` is out of place here, and I don't see any reason for this kind of message to be written to `stdout`.
I suggest we remove this.

PR #216, opened 2017-06-10
lib: remove console.warn() from safeRequire()

DzyubSpirit:
Refs: #188


DzyubSpirit:
I think if user wants to get error he will use `try... catch`. 
I based my opinion on single usage of that function.  And that usage shows that `safeRequire` is designed for

DzyubSpirit:
@aqrln Could you please say if that implementation of `safeRequire` suits the user needs?

nechaido:
@DzyubSpirit I'd prefer `safeReqire()` to return tuple `[eroor, module]` instead.

lib/common.js
```diff
@@ -31,14 +31,77 @@ common.forwardMultipleEvents = (from, to, eventNames) => {
...
 
 // This function can be used in contexts where a function (e.g., a callback) is
 // required but no actions have to be done.
 //
 common.doNothing = () => {};
+
+// Either can be used for saving either left type or right type value.
+// Left type is commonly used for additional information like error.
+//
+function Either(left, right) {
```
aqrln: This doesn't belong to the scope of this package.

package.json
```diff
@@ -34,7 +34,8 @@
     "./lib/simple-auth.js": false
   },
   "dependencies": {
-    "websocket": "^1.0.24"
+    "websocket": "^1.0.24",
+    "metarhia-common": "^0.0.9"
```
aqrln: This package won't be installed at all because npm will just use `package-lock.json` if it is present (and it is). Please make sure that you use npm5.

aqrln:
@DzyubSpirit ping. Are you still up to working on this?

DzyubSpirit:
@aqrln Yes, there is metarhia/common#31 about Either type because Either doesn't belong to the scope of this package. Also I've changed package-lock.json but forget to push, thank you for remind.

lib/record-serialization.js
```diff
@@ -12,20 +13,23 @@ module.exports = jsrs;
 // one of our priorities to optimize it.
 const USE_NATIVE_SERIALIZER = false;
 
-const jstpNative =
-  safeRequire('../build/Release/jstp') ||
-  safeRequire('../build/Debug/jstp');
+const jstpNativeE = altConcatLazy(
```
tshemsedinov: We have `require(name)`
We need `api.common.require(name1, ..., nameN)`
We can write `api.common.either(fn)` that which will expect `fn` is synchronous an it may raise an `Error`
If all calls to `fn` will raise, resulted function will raise too.
So we can use it like this:
```js
api.common.require = api.common.either(require);
const moduleName = api.common.require(name1, name2, name3);
```

aqrln:
Alternative PR: https://github.com/metarhia/jstp/pull/226. Whichever wins the race will land on master.

nechaido:
Closed in favour of https://github.com/metarhia/jstp/pull/226.

Issue #152, opened 2017-05-10
Make the parser and serializer compliant with JSON5 specs


aqrln:
Not compliant as in "strictly adhering to" but as in "must parse everything that JSON5 allows and probably beyond and must serialize to valid JSON5".

belochub:
@metarhia/jstp-core, I've updated the description to represent current progress on this issue.

PR #218, opened 2017-06-12
parser: parse Unicode identifiers

belochub:


belochub:
@aqrln, may you help me with creation of pre-build configuration script to make it possible to choose which of the two implementations to use (check out https://github.com/metarhia/jstp/commit/89427481f29d1c50b2cd6ec042493a251cef26cf#diff-9689e209b75f216133f2b377028d89f6R9)?

tools/make-unicode-tables.js
```diff
@@ -0,0 +1,186 @@
...
+const UCD_LINK = 'http://www.unicode.org/Public/' + UNICODE_VERSION +
+  '/ucd/DerivedCoreProperties.txt';
+const fullTablesFilename = 'unicode_tables.h';
+const rangeTablesFilename = 'unicode_range_tables.h';
+const getHeaderGuard = filename =>
+  `SRC_${filename.replace(/\W/g, '_').toUpperCase()}_`;
+const getOutputPath = filename => path.join(__dirname, '../src', filename);
+
+const getFileHeader = filename =>
+`// Copyright (c) 2017 JSTP project authors. Use of this source code is
```
lundibundi: Maybe it'll be better to put license in a separate file?

belochub: In my opinion it is good to have it here, because it will probably help to avoid any legal problems, given the fact that this file is also a piece of software which works with Unicode data files. Consider this as a precaution, I don't know for sure but I also don't want to find out about it in a bad way.

It helps to understand the resulting format of the output files as well, so that you don't need to look anywhere else or open this auto-generated files (which are pretty big and can slow down some of the editors) to understand their contents.

lundibundi: Fine with me =), reasonable enough.

aqrln:
@belochub can you please rebase to resolve conflicts?

tools/make-unicode-tables.js
```diff
@@ -0,0 +1,186 @@
...
+const idContinueRanges = [];
+
+let idStartTotalCount = 0;
+let idContinueTotalCount = 0;
+
+const lineRegex = /^([0-9A-F]{4,6})(?:\.\.([0-9A-F]{4,6}))? *; (\w+) #.*$/;
+
+http.get(UCD_LINK, (res) => {
+  const linereader = byline.createStream(res);
+  linereader.on('data', (line) => {
```
aqrln: Well, wouldn't the core `readline` module be sufficient to accomplish this?

belochub: Yeah, it probably is sufficient, I missed that.

package-lock.json
```diff
@@ -4637,8 +4637,7 @@
             "wordwrap": {
               "version": "0.0.3",
               "resolved": "https://registry.npmjs.org/wordwrap/-/wordwrap-0.0.3.tgz",
-              "integrity": "sha1-o9XabNXAvAAI03I0u68b7WMFkQc=",
-              "dev": true
```
aqrln: Huh? Looks like npm correctly handles the case when a package becomes used in `dependencies` and not only in `devDependencies`, but does not handle the case when the package is not required for `dependencies` anymore. Can you please revert this change manually?

aqrln: Oh, wait, I think you only modified `devDependencies`. Now I really wonder what happened here.

belochub:
@aqrln, we can't land it yet, you still haven't addressed https://github.com/metarhia/jstp/pull/218#issuecomment-307757110.

belochub:
@aqrln, ping.

aqrln:
@belochub I've pushed a commit to your branch.
@nechaido @lundibundi PTAL

aqrln:
Landed in 4e942ae06d0a291bea37a6c6ffb42642b9842d4f and 4b67cbdb2a50266fa5f6ab6f1760811fc29465e0.

PR #219, opened 2017-06-12
parser: parse Unicode escape sequences in keys

belochub:
Also, add support for Unicode surrogate pairs and improve key parsing performance by avoiding copying when possible.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -554,6 +554,57 @@ MaybeLocal<Value> ParseString(Isolate*    isolate,
 
 static unsigned int ReadHexNumber(const char* str, size_t len, bool* ok);
 
+// Parses a Unicode escape sequence after the '\u' part and returns it's
+// code point value. Supports surrogate pairs. Total size of escape
+// sequence (excluding first '\u') is written in `size`.
+static uint32_t ReadUnicodeEscapeSequence(Isolate* isolate,
+                                          const char* str,
+                                          size_t* size,
+                                          bool* ok) {
+  uint32_t result = 0xFFFD;
+
+  if (isxdigit(str[0])) {
+    result = ReadHexNumber(str, 4, ok);
+    *size = 4;
+  } else if (str[0] == '{') {
+    size_t hex_size;
+    for (hex_size = 1;
+         str[hex_size + 1] != '}';
+         hex_size++) {
+      if (str[hex_size + 1] == '\0') {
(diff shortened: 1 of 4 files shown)


aqrln:
@belochub can you please rebase it? I think removing the old `package-lock.json`, resolving the conflict in `package.json` and regenerating the package-lock should be just fine.

nechaido:
@belochub can you, please, rebase this PR on master?

belochub:
@aqrln, @lundibundi, ping.

src/jsrs_parser.cc
```diff
@@ -554,6 +555,57 @@ MaybeLocal<Value> ParseString(Isolate*    isolate,
...
+// sequence (excluding first '\u') is written in `size`.
+static uint32_t ReadUnicodeEscapeSequence(Isolate* isolate,
+                                          const char* str,
+                                          size_t* size,
+                                          bool* ok) {
+  uint32_t result = 0xFFFD;
+
+  if (isxdigit(str[0])) {
+    result = ReadHexNumber(str, 4, ok);
+    *size = 4;
```
aqrln: Should it check whether `*ok` is true before changing `*size`?

src/jsrs_parser.cc
```diff
@@ -554,6 +555,57 @@ MaybeLocal<Value> ParseString(Isolate*    isolate,
...
+    for (hex_size = 1;
+         str[hex_size + 1] != '}';
+         hex_size++) {
+      if (str[hex_size + 1] == '\0') {
+        THROW_EXCEPTION(SyntaxError, "Invalid Unicode code point escape");
+        *ok = false;
+        return 0xFFFD;
+      }
+    }
+    result = ReadHexNumber(str + 1, hex_size, ok);
```
aqrln: It should check whether `result` is a valid code point (i.e., something like `\u{1234123412341234}` must be a syntax error).

belochub: Yeah, I looked into it and decided to reimplement `ReadHexNumber` overall, because it has some other problems.

belochub:
@aqrln, I've updated `ReadHexNumber` function to make it safer and added tests for it, PTAL.
Also, this probably fixed parsing of strings like this: `'\u00001'` with the digits after the escape sequence, this case was incorrectly considered an error before.

aqrln:
Landed in bef3572c2967d4da668d92e7d944efd58ca3f15b.

PR #220, opened 2017-06-12
parser: improve string parsing performance

belochub:
Improve string parsing performance by avoiding copying when possible.

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -462,11 +462,10 @@ MaybeLocal<Value> ParseString(Isolate*    isolate,
     }
 
     if (begin[i] == '\\') {
+      use_fallback(i - 1);
       if (IsLineTerminatorSequence(begin + i + 1, &in_offset)) {
-        use_fallback(i - 1);
         i += in_offset;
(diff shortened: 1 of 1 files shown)


src/jsrs_parser.cc
```diff
@@ -456,8 +463,10 @@ MaybeLocal<Value> ParseString(Isolate*    isolate,
 
     if (begin[i] == '\\') {
       if (IsLineTerminatorSequence(begin + i + 1, &in_offset)) {
+        use_fallback(i - 1);
```
aqrln: We can move this line to the outer block and get rid of the analogous one in the `else` clause, can't we?

belochub: @aqrln
Nope, it won't work the same.

belochub: @aqrln 
Yeah, after looking at it one more time it looks okay, I changed it like you proposed.

aqrln:
Latest changes LGTM. Landed in 1f1d7d0a094a0785c8b5ebd144ca41e658e6cc7a.

PR #221, opened 2017-06-12
parser: parse noctal number literals

belochub:

### DIFF
--- a/src/jsrs_parser.cc
+++ b/src/jsrs_parser.cc
@@ -290,17 +290,15 @@ MaybeLocal<Value> ParseBool(Isolate*    isolate,
   return result;
 }
 
-// Checks if a character is an octal digit.
-inline bool IsOctalDigit(char character) {
-  return character >= '0' && character <= '7';
-}
-
 MaybeLocal<Value> ParseNumber(Isolate*    isolate,
                               const char* begin,
                               const char* end,
                               size_t*     size) {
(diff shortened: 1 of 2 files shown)


src/jsrs_parser.cc
```diff
@@ -426,6 +425,55 @@ Local<Value> ParseBigIntegerNumber(Isolate*    isolate,
...
+      break;
+    }
+
+    current_digit = begin[i] - '0';
+    if (current_digit > 7) {
+      is_octal = false;
+    }
+
+    if (!int32_overflow) {
+      result *= 10;
```
nechaido: Shouldn't it be 8?

src/jsrs_parser.cc
```diff
@@ -426,6 +425,55 @@ Local<Value> ParseBigIntegerNumber(Isolate*    isolate,
...
+      result += current_digit;
+    } else {
+      overflow_result *= 10.0;
```
nechaido: Ditto.

src/jsrs_parser.cc
belochub: No, because then it will be octal number parsing.

nechaido: Oh, yep. Ok then.

aqrln:
Landed in 726c8d8f7adad04170722f3396a21e7162920f39.

Issue #49, opened 2017-02-06
Optimize Connection events


aqrln:
* Review existing events, remove unnecessary ones and rename long ones.
* Prevent possible deopts in `Connection#_emitPacketEvent` or even get rid of this method completely.

PR #222, opened 2017-06-16
lib: optimize connection events

lundibundi:
* remove unnecessary events like 'handshake', 'call', 'handshakeRequest'
* replace 'callbackForUnknownPacket' with rejectPacket
* change signature of connection 'event' (remote event), now it contains
  interfaceName, eventName, eventArgs

Also event though state packet event is now removed it has to be put back with proper data inside the event, because _emitPacketEvent it used adds a lot of redundant data. Can you please comment on which data should be included in `state` event?
UPD: as discused state packets are removed for now.

Fixes: https://github.com/metarhia/jstp/issues/49

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -58,6 +58,8 @@ function Connection(transport, server, client) {
 
   this._heartbeatCallbackInstance = null;
 
+  // Defined in constructor to be used as default callback in callMethod
+  // without binding it.
   this._emitError = (error) => {
     if (error) this.emit('error', error);
   };
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -58,6 +58,10 @@ function Connection(transport, server, client) {
 
   this._heartbeatCallbackInstance = null;
 
+  this._emitError = (error) => {
+    if (error) this.emit('error', error);
+  };
+
   transport.on('packet', this._processPacket.bind(this));
   transport.on('close', this._onSocketClose.bind(this));
   transport.on('error', this._onSocketError.bind(this));
@@ -77,9 +81,7 @@ Connection.prototype.callMethod = function(
 ) {
   const packet = this.createPacket('call', interfaceName, methodName, args);
   const packetId = packet.call[0];
-  this._callbacks[packetId] = callback || ((error) => {
-    if (error) this.emit('error', error);
-  });
+  this._callbacks[packetId] = callback || this._emitError;
   this._send(packet);
 };
 
@@ -393,12 +395,6 @@ Connection.prototype._processHandshakeRequest = function(packet, keys) {
   const credentials = authStrategy && packet[authStrategy];
   authStrategy = authStrategy || 'anonymous';
 
-  this._emitPacketEvent('handshakeRequest', packet, packet.handshake[0], {
-    packetType: 'handshake',
-    handshakeRequest: true,
-    authStrategy
-  });
-
   this.server.startSession(this, application, authStrategy, credentials,
     this._onSessionCreated.bind(this));
 };
@@ -442,10 +438,6 @@ Connection.prototype._processHandshakeResponse = function(packet) {
     this.handshakeDone = true;
     this.application = this.client.application;
 
-    this._emitPacketEvent('handshake', packet, packetId, {
-      sessionId: packet.ok
-    });
-
     callback(null, packet.ok);
   } else if (packet.error) {
     delete this._callbacks[packetId];
@@ -475,11 +467,6 @@ Connection.prototype._processCallPacket = function(packet, keys) {
   const methodName = keys[1];
   const args = packet[methodName];
 
-  this._emitPacketEvent('call', packet, packetId, {
-    interfaceName,
-    methodName
-  });
-
   const callback = this._remoteCallbackWrapper.bind(this, packetId);
 
   if (!args) {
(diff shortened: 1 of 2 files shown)


belochub:
I think that it is better to add logger functionality in a different PR or at least in a different commit in this PR because it isn't really connected to connection events optimization.

belochub:
Also, it would be great to have an ability to add multiple loggers instead of setting just one.

lundibundi:
@belochub yeah, different commit I guess even though the change is good enough to be in a different PR I  think there is no need in that because it will not use it anywhere, so even if you set the logger it will do nothing, but having a different PR is also fine with me.

Could you elaborate on the latter comment, what multiple loggers can you have?

belochub:
I meant that you can create a new PR based on this PR with logger functionality.

About multiple loggers, there is a lot of possible variants, for example, one logger for console output, another for file logging and the third for logging over the network. Of course, it can be implemented even with the current changes, but it will be much easier if you add a possibility to add multiple loggers.

lundibundi:
@belochub I do not see any reason to implement it here, I think that's the job of a logging framework user decides to use with our project, the logger I add here is just an api (at least [winston](https://github.com/winstonjs/winston#working-with-multiple-loggers-in-winston) and [bunyan](https://github.com/trentm/node-bunyan#streams-introduction) can do that pretty good).

belochub:
Yeah, but it will do no harm if you add the possibility to have multiple loggers. It will make it possible to use small and simple self-written loggers instead of bigger frameworks, and you can still use the frameworks by adding just one logger instance.

lundibundi:
As discussed, waiting for https://github.com/metarhia/jstp/pull/224.

lundibundi:
Properly split changes, removing blocked.

lib/connection.js
```diff
@@ -58,6 +58,10 @@ function Connection(transport, server, client) {
 
   this._heartbeatCallbackInstance = null;
 
+  this._emitError = (error) => {
```
aqrln: Can you please add a comment explaining the reason of this being created in the constructor instead of on the prototype? It is not immediately obvious until one reads the `callMethod()` code.

test/node/connection-event.js
```diff
@@ -42,16 +42,17 @@ test.test('server must process an event', (test) => {
   client.connectAndHandshake(app.name, null, null, (error, connection) => {
     test.assertNot(error, 'must connect to server');
 
-    server.getClients()[0].on('event', (event) => {
-      test.strictEqual(event.interfaceName, iface,
+    server.getClients()[0].on('event',
+      (interfaceName, remoteName, remoteArgs) => {
+        test.strictEqual(interfaceName, iface,
         'event interface must match');
```
aqrln: Nit: can you please indent this line?

test/node/connection-event.js
```diff
@@ -42,16 +42,17 @@ test.test('server must process an event', (test) => {
...
-      test.strictEqual(event.remoteEventName, eventName,
+        test.strictEqual(remoteName, eventName,
         'event name must be equal to the emitted one');
```
aqrln: Ditto.

test/node/connection-event.js
```diff
@@ -42,16 +42,17 @@ test.test('server must process an event', (test) => {
...
-      test.strictDeepEqual(event.remoteEventArgs, args,
+        test.strictDeepEqual(remoteArgs, args,
         'event arguments must be equal to the passed ones');
```
aqrln: Ditto.

test/node/connection-event.js
```diff
@@ -80,15 +81,16 @@ test.test('remote proxy must emit an event', (test) => {
     (error, connection, api) => {
       test.assertNot(error, 'must connect to server');
 
-      server.getClients()[0].on('event', (event) => {
-        test.strictEqual(event.interfaceName, iface,
+      server.getClients()[0].on('event',
+        (interfaceName, remoteName, remoteArgs) => {
+          test.strictEqual(interfaceName, iface,
           'event interface must match');
```
aqrln: Ditto.

test/node/connection-event.js
```diff
@@ -80,15 +81,16 @@ test.test('remote proxy must emit an event', (test) => {
...
-        test.strictEqual(event.remoteEventName, eventName,
+          test.strictEqual(remoteName, eventName,
           'event name must be equal to the emitted one');
```
aqrln: Ditto.

test/node/connection-event.js
```diff
@@ -80,15 +81,16 @@ test.test('remote proxy must emit an event', (test) => {
...
-        test.strictDeepEqual(event.remoteEventArgs, args,
+          test.strictDeepEqual(remoteArgs, args,
           'event arguments must be equal to the passed ones');
```
aqrln: Ditto.

lundibundi:
@belochub it also reuses same error callback, removes unneded data from jstp `event` and replaces `callbackForUnknownPacket` with 'rejectPacket', should we really just call this one 'remove redundant events'?

test/node/connection-event.js
lundibundi: Sure. We should find a way to track it automatically =).

aqrln: It will be tracked automatically as soon as we update ESLint to 4.0. The only reason I haven't done it yet is that we use the `eslint-plugin-import` plugin which still has `eslint@3.x` as its peerDependency (though it seems to work with the latest eslint nevertheless).

belochub:
@lundibundi, I didn't say that we should call it "remove redundant events".

lundibundi:
@belochub I just suggested a name that will present that this PR removes those events, what is your suggestion?

belochub:
@lundibundi, I am sorry, I misunderstood your comment, to me it looked like you disagree with the idea of renaming this at all.
In my opinion, a better name for this could be "avoid emitting unused events", but it isn't perfect as well.

aqrln:
The current commit message is fine since the PR does not only vanish redundant events, but also performs various other performance-related optimization (e.g., avoiding creation of extra closure in order to emit `'error'` event in `callMethod()`) and the removal of unnecessary events is explicitly stated in the full commit message.

aqrln:
Landed in 637bb40730c555bd90a45a98f7b5b18eca01c060.

aqrln:
@belochub please let me know if I was too fast on this and I should have waited for your reply.

PR #223, opened 2017-06-17
parser: allow numeric literals as object keys

belochub:


belochub:
@nechaido, ping.

belochub:
@nechaido, adding the test you proposed to `/test/fixtures/serde-test-cases/serde.js`, will definitely break the build, this doesn't belong to regular test cases yet, in my opinion, it should be added to current json5 tests because it belongs to json5 specification and it isn't implemented in JavaScript parser yet, which makes it a "to do" test case for now.
I will add the corresponding test to json5 test cases if you don't mind.

nechaido:
@belochub you are right, go ahead.

belochub:
@nechaido, I've added the test, PTAL.

aqrln:
Landed in 635ea15ccbfa540e4d9b84dda26ab40561d487ca.

Issue #102, opened 2017-03-13
Support different loggers


aqrln:
Instead of writing errors and warnings to stderr, make loggers pluggable.

Ref: https://github.com/metarhia/JSTP/pull/101#discussion_r105788964

tshemsedinov:
@lundibundi @aqrln paused, I'll give comments ASAP about code in #224

tshemsedinov:
Impress logger is an object having a set of logging methods with single string argument:
`logger.method(message: string)`

By default we have following methods:
- to log errors and debugging: `error, debug, warning`
- to log API requests: `access, api, slow`
- to log server events: `server, node, cloud`

But application can have additional loggers it is configurable: `config/log.js`

JSTP may use `access, api, slow` to log RPC access. Logger will create separate file per process (to have exclusive write). But `server` and `cloud` logs have single file per server (no separate file per process). Impress logger implements log rotation (new file each day and delete old files).

We do not need:
- support winston or bunyan
- logging level (number)
- `.isDebug()` or `.isWarn()`
  - you can do something like this `const debug = logger.debug || () => {};` instead

@metarhia/jstp-core

belochub:
@tshemsedinov, `.isDebug()` and other similar methods are needed to avoid unnecessary preparation in cases when there is no such method in the provided logger (see https://github.com/metarhia/jstp/pull/224#issuecomment-310655557).

tshemsedinov:
`if (logger.debug) ...` @belochub

tshemsedinov:
Even more, you do not need to know whether app is in debug mode, just call `logger.debug(message)` and it will be logged to file or to console or both or ignored at all. @belochub

belochub:
It doesn't check whether an application is in debug mode or not, it's just checking if the logger is attached to avoid preparation (like serialization or removing sensitive information, see https://github.com/metarhia/jstp/pull/224#issuecomment-310655557).
Even if there is a function which ignores the message, it is better to avoid calling it, because otherwise, additional message processing is unavoidable and this will harm the overall performance.

PR #224, opened 2017-06-17
lib: support different loggers

lundibundi:
Enables users to specify logger they want to use with our library. Also adds simple proxy to `console`.

As discussed in https://github.com/metarhia/jstp/issues/102 I extract logger fuctionality to a different PR.

Also I'd like to add a few tests for console logger, but I don't know how to better implement it, as I don't think that either redirecting console output during the test or just checking stdout is a good solution, can someone suggest a solution?

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -2,18 +2,60 @@
 
 let logger = null;
 
-const set = l => logger = l;
+const defaultLevelCallback = level => logger && logger[level];
+let isLevel = defaultLevelCallback;
+
+const levelToNumber = {
+  fatal: 60, error: 50,
+  warn: 40, info: 30,
+  verbose: 25, debug: 20,
+  silly: 10, trace: 10
+};
+
+const winstonLevelCallback =
+  level => levelToNumber[logger.level] <= levelToNumber[level];
+
+const bunyanLevelCallback =
+  level => logger.level() <= levelToNumber[level];
+
+// Set logger that will be used to forward JSTP logging
+//   loggerInstance - instance of logger that may have methods
+//                    'debug', 'verbose', 'info', 'warn', 'error', 'silly'.
+//                    It must have levels that levelCallback
+//                    reports as existing.
+//   levelCallback - function to check current logging level, will be
+//                   called with one of the:
+//                   'debug', 'verbose', 'info', 'warn', 'error', 'silly'.
+//                   by default defaultLevelCallback will be used that only
+//                   checks that such method exists.
+//                   You can provide your own or use one of the available:
+//                   winston (winstonLevelCallback) or
+//                   bunyan(bunyanLevelCallback).
+//                   In case of bunyan to support silly logging level
+//                   you must map 'silly' function to 'trace'.
+//
+const set = (loggerInstance, levelCallback) => {
+  if (loggerInstance) logger = loggerInstance;
+  if (levelCallback) isLevel = levelCallback;
+};
+
 const get = () => logger;
 
-const isDebug = () => logger && logger.debug;
-const isVerbose = () => logger && logger.verbose;
-const isInfo = () => logger && logger.info;
-const isWarn = () => logger && logger.warn;
-const isError = () => logger && logger.error;
+const isSilly = () => isLevel('silly');
+const isDebug = () => isLevel('debug');
+const isVerbose = () => isLevel('verbose');
+const isInfo = () => isLevel('info');
+const isWarn = () => isLevel('warn');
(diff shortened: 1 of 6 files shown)


lib/logger.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+let logger = null;
+
+const set = l => logger = l;
+const get = () => logger;
+
+const isDebug = () => logger && logger.debug;
```
aqrln: I find these names misleading, something like `hasDebug` etc would be better, IMO. I'd expect `isDebug`, `isVerbose` and alike to report the current logging level, which is not the case with these checks because these methods will be present in virtually every logger regardless of the logging level.

lib/record-serialization.js
```diff
@@ -22,10 +23,12 @@ if (jstpNative) {
     jsrs.stringify = serialize;
   }
 } else {
-  console.warn(
-    'JSTP native addon is not built or is not functional. ' +
-    'Run `npm install` in order to build it, otherwise you will get ' +
-    'poor performance.'
-  );
+  if (logger.isWarn()) {
```
aqrln: It isn't possible to replace the logger before this statement will be executed, so it will always use the default logger.

lundibundi: @aqrln yeah, missed that. Should we set `console` as default logger or always log to console here?

lib/logger.js
lundibundi: That was the actual reason I named them as such, I was mislead by [issue](https://github.com/winstonjs/winston/issues/215) in `winston` that actually suggested to check level this way.
I haven't found a good way to get debug level out of winston beside `.level` which gives a string, that's disappointing. Lucky enough bunyan returns a number which is a lot better. I guess we will allow user to provide a callback to check current log level and I'll write proxies for winston and bunyan to simplify their usage.

nechaido:
I'd like the logger to have method's like: `warn`, `error`, `log`, `debug`. Logger should decide for itself to log something or not, that way we'll have no need to write this kind of statements:
`if (logger.isDebug()) logPacket(packet);`.

lundibundi:
@nechaido It already has those methods and decides by itself whether he should log or drop the message. The need in methods like `isDebug()` and such arises from the cost that preparation for logging has, for example in this PR I have to check jstp message type and copy if necessary while removing some fields to avoid leak of private data to logs (for handshake packet). Checking `isDebug()` is much cheaper than that. And also I don't think that implementing callback to check current logging level is that hard.

nechaido:
@lundibundi this could be achieved by passing a function to logger which will do the preparation and then return the string to be logged.

belochub:
@lundibundi, can't you make a wrapper in our library which will check the logger level before the preparation, to avoid boilerplate checking like the one @nechaido mentioned?

lundibundi:
@belochub I can do that but I see no reason for that complexity, whether we write

nechaido:
@lundibundi and what about:
```
logger.debug(() => preparePacket(packet));
```
Looks ok for me ~~and doesn't do any unnecessary preparation in release mod, only call to empty function.~~

Forgot about creating a function to be passed.

belochub:
@lundibundi, what you said in https://github.com/metarhia/jstp/pull/224#issuecomment-310670911 wasn't at all what I proposed to do.

belochub:
@lundibundi, what I proposed will make
```js
if (logger.isDebug()) logPacket(packet);
```
look like
```js
logger.get().debug(packet);
```
or even
```js
logger.debug(packet);
```

lundibundi:
@belochub oh, sorry, I thought that you extended @nechaido idea. So you want a logger specifically to log jstp messages, that will check log level and if object is passed treat it as jstp message and log it same way `logPacket` does and if string is passed log it as is, is my understanding correct?

belochub:
@lundibundi, yeah, something like that.

lundibundi:
@belochub but its only usage are those 3 calls in connection.js and I'm not sure if it is plausible to implement it solely for those usages? Also I can move `isDebug` check into `logPacket` so there'll be only one check for `isDebug()` (in code), they only reason I left it outside was to avoid unnecessary call to `logPacket`.

aqrln:
@lundibundi ping

lundibundi:
@aqrln I'm waiting for the decision of our team, whether this is a good solution or not. We even got the slack vote which has apparently been flooded by github bot.

belochub:
Closed in favor of https://github.com/metarhia/jstp/pull/252 (see https://github.com/metarhia/jstp/issues/250 for more information).

PR #225, opened 2017-06-19
deps: update deps and regenerate package-lock.json

aqrln:
Using the current `package-lock.json` (that was generated using
npm@5.0.0 and then amended a couple of times with newer npm versions),
npm generates a dependency tree that shows lots of errors in the `npm
list` output. This commit regenerates the package tree with the latest
npm version in order to resolve these problems.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,7 +4,8 @@
   "lockfileVersion": 1,
   "dependencies": {
     "acorn": {
-      "version": "https://registry.npmjs.org/acorn/-/acorn-5.0.3.tgz",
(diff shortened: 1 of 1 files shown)


aqrln:
Rebased to incorporate new package from `master`. Additionally, since I regenerated the package-lock using `rm -r node_modules package-lock.json && npm i`, many dependency have been bumped, so I renamed both the commit and PR.

aqrln:
Landed in 54954b7731675338a5857d8eddf925587ef03019.

PR #226, opened 2017-06-20
lib: make `safeRequire()` return tuple

### LINKED ISSUE #188: Remove unnecessary output from common.safeRequire()

nechaido:
As we discussed in #216 I add a PR that fixes #188 using different approach.
Refs:#216.
Closes: #188.

### DIFF
--- a/lib/record-serialization.js
+++ b/lib/record-serialization.js
@@ -12,7 +12,7 @@ module.exports = jsrs;
 // one of our priorities to optimize it.
 const USE_NATIVE_SERIALIZER = false;
 
-const jstpNative =
+const [error, jstpNative] =
   safeRequire('../build/Release/jstp') ||
   safeRequire('../build/Debug/jstp');
 
@@ -23,6 +23,7 @@ if (jstpNative) {
   }
 } else {
   console.warn(
+    error + '\n' +
     'JSTP native addon is not built or is not functional. ' +
     'Run `npm install` in order to build it, otherwise you will get ' +
     'poor performance.'
--- a/lib/record-serialization.js
+++ b/lib/record-serialization.js
@@ -12,9 +12,11 @@ module.exports = jsrs;
 // one of our priorities to optimize it.
 const USE_NATIVE_SERIALIZER = false;
 
-const [error, jstpNative] =
-  safeRequire('../build/Release/jstp') ||
-  safeRequire('../build/Debug/jstp');
+let [error, jstpNative] = safeRequire('../build/Release/jstp');
+
+if (error) {
+  [error, jstpNative] = safeRequire('../build/Debug/jstp');
+}
 
 if (jstpNative) {
   Object.assign(jsrs, jstpNative);
--- a/lib/record-serialization.js
+++ b/lib/record-serialization.js
@@ -15,6 +15,7 @@ const USE_NATIVE_SERIALIZER = false;
 let [error, jstpNative] = safeRequire('../build/Release/jstp');
 
 if (error) {
+  console.warn(error.toString());
   [error, jstpNative] = safeRequire('../build/Debug/jstp');
 }
 
--- a/test/node/common-safe-require.js
+++ b/test/node/common-safe-require.js
@@ -7,7 +7,7 @@ const common = require('../../lib/common');
 const existingModule = 'fs';
 const nonExistingModule = '__non_existing_module__';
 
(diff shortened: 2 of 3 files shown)


lib/record-serialization.js
```diff
@@ -12,7 +12,7 @@ module.exports = jsrs;
 // one of our priorities to optimize it.
 const USE_NATIVE_SERIALIZER = false;
 
-const jstpNative =
+const [error, jstpNative] =
   safeRequire('../build/Release/jstp') ||
```
aqrln: This won't work.

test/node/common-safe-require.js
```diff
@@ -7,7 +7,7 @@ const common = require('../../lib/common');
 const existingModule = 'fs';
 const nonExistingModule = '__non_existing_module__';
 
-test.ok(common.safeRequire(existingModule), 'must require existing module');
+test.ok(common.safeRequire(existingModule), 'must require existing module'[1]);
 
-test.equal(common.safeRequire(nonExistingModule), null,
+test.equal(common.safeRequire(nonExistingModule)[1], null,
```
aqrln: `[0]`

test/node/common-safe-require.js
```diff
@@ -7,7 +7,7 @@ const common = require('../../lib/common');
 const existingModule = 'fs';
 const nonExistingModule = '__non_existing_module__';
 
-test.ok(common.safeRequire(existingModule), 'must require existing module');
+test.ok(common.safeRequire(existingModule), 'must require existing module'[1]);
```
aqrln: I believe you didn't mean to compare an array with 'm' ;)

lundibundi: I think `[1]` should be in a different place =).

test/node/common-safe-require.js
lundibundi: Why don't we test this with `notOk` as the one above is tested with `ok`?

test/node/common-safe-require.js
lundibundi: That's error message =). But yeah it was funny.

aqrln: ah, it's not a comparison, but a test message

lib/record-serialization.js
```diff
@@ -12,9 +12,11 @@ module.exports = jsrs;
 // one of our priorities to optimize it.
 const USE_NATIVE_SERIALIZER = false;
 
-const jstpNative =
-  safeRequire('../build/Release/jstp') ||
-  safeRequire('../build/Debug/jstp');
+let [error, jstpNative] = safeRequire('../build/Release/jstp');
+
+if (error) {
+  [error, jstpNative] = safeRequire('../build/Debug/jstp');
```
belochub: `error` will be overwritten here, losing the original error.

aqrln: Yeah, I'd rather only show the error that occurred trying to load the release build when both failed if we should only show only one. Alternatively, we may show both, like we do now. I'm fine with it both ways, slightly leaning towards the former one.

aqrln:
Alternative PR: https://github.com/metarhia/jstp/pull/216. Whichever wins the race will land on master.

nechaido:
Landed in 35c6a8cbccfe7e91e5d4a9dce30df371b7448c06.

Issue #31, opened 2016-12-27
API versioning


aqrln:
It must be possible to provide several versions of an API and to select one during a handshake.

tshemsedinov:
Просто `interfaceName + '.' + version`

PR #231, opened 2017-06-22
lib: implement API versioning

lundibundi:
* allow to specify API version during handshake
* make createAppsIndex handle multiple API versions
* add API version parameter to Application
* add API version parameter to SimpleConnectPolicy
* add tests for interface versioning

Fixes: #31.

### DIFF
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -2,6 +2,7 @@
 
 const semver = require('semver');
 
+const common = require('./common');
 const errors = require('./errors');
 
 // Generic application class. You are free to substitute it with whatever suits
@@ -16,7 +17,7 @@ const errors = require('./errors');
 //
 class Application {
   constructor(name, api, version) {
-    [this.name, this.version] = name.split('@');
+    [this.name, this.version] = common.rsplit(name, '@');
     if (!this.version) this.version = version;
     if (this.version && !semver.valid(this.version)) {
       throw new TypeError('Invalid semver version');
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -1,15 +1,26 @@
 'use strict';
 
+const semver = require('semver');
+
 const errors = require('./errors');
 
 // Generic application class. You are free to substitute it with whatever suits
 // your needs.
-//   name - application name
-//   api - application API
+//   name - application name that may contain version after '@'
+//          (e.g. app@1.0.0). Version in name is preferred over
+//          'version' parameter
+//   api - object that contains interfaces as its fields each of which
+//         contains functions by method names. Each method has the following
+//         signature (connection, <0 or more method arguments>, callback)
+//   version - application version of 'name' application
 //
 class Application {
-  constructor(name, api) {
-    this.name = name;
+  constructor(name, api, version) {
+    [this.name, this.version] = name.split('@');
+    if (!this.version) this.version = version;
+    if (this.version && !semver.valid(this.version)) {
+      throw new TypeError('Invalid semver version');
+    }
     this.api = api;
   }
 
@@ -54,10 +65,41 @@ class Application {
 //   applications - array of JSTP applications
 //
 const createAppsIndex = (applications) => {
-  const index = {};
+  const latestApps = new Map();
+  const index = new Map();
 
   applications.forEach((application) => {
-    index[application.name] = application;
+    let appVersions = index.get(application.name);
+    if (!appVersions) {
+      appVersions = new Map();
+      index.set(application.name, appVersions);
+    }
+    if (!application.version) {
+      // no version means latest version
+      if (appVersions.has('latest')) {
+        throw new Error(
+          `Multiple entries of '${application.name} without version`
+        );
+      } else {
+        appVersions.set('latest', application);
+      }
+    } else {
+      appVersions.set(application.version, application);
+
+      // save latest version to fill missing latest versions later
+      const latestApp = latestApps.get(application.name);
+      if (!latestApp || semver.gt(application.version, latestApp.version)) {
+        latestApps.set(application.name, application);
+      }
+    }
+  });
+
+  // set latest versions of apps without explicit latest version
+  latestApps.forEach((application, appName) => {
+    const appVersions = index.get(appName);
+    if (!appVersions.has('latest')) {
+      appVersions.set('latest', application);
+    }
   });
 
   return index;
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const { EventEmitter } = require('events');
+const semver = require('semver');
 const timers = require('timers');
 
 const common = require('./common');
@@ -117,15 +118,46 @@ class Connection extends EventEmitter {
   }
 
   // Send a handshake message over the connection
-  //   appName - name of an application to connect to
+  //   app - string or object, application to connect to as 'name' or
+  //         'name@version' or { name, version }, where version
+  //         must be a valid semver range
   //   login - user name (optional)
   //   password - user password (optional)
   //   callback - callback function to invoke after the handshake is completed
   //
-  handshake(appName, login, password, callback) {
-    const message = login && password ?
-      this._createMessage('handshake', appName, 'login', [login, password]) :
-      this._createMessage('handshake', appName);
+  handshake(app, login, password, callback) {
+    let name, version;
+    if (typeof app === 'string') {
+      [name, version] = app.split('@');
+    } else {
+      name = app.name;
+      version = app.version;
+    }
+
+    if (version && !semver.validRange(version)) {
+      const error = new Error('Invalid semver version range');
+      if (callback) {
+        callback(error);
+      } else {
+        this.emit('error', error);
+      }
+      return;
+    }
+
+    let message;
+    if (version) {
+      message = login && password ?
+        this._createPacketWithArray(
+          'handshake', [name, version], 'login', [login, password]
+        ) :
+        this._createPacketWithArray(
+          'handshake', [name, version]
+        );
+    } else {
+      message = login && password ?
+        this._createMessage('handshake', name, 'login', [login, password]) :
+        this._createMessage('handshake', name);
+    }
 
     const messageId = message.handshake[0];
     this._callbacks[messageId] = (error, sessionId) => {
@@ -214,6 +246,28 @@ class Connection extends EventEmitter {
       this.clearTimeout(this._heartbeatCallbackInstance);
       this._heartbeatCallbackInstance = null;
     }
+  }
+
+  // Create a JSTP message
+  //   kind - message kind
+  //   target - array of arguments for kind key, usually a name
+  //            of an interface or an application (optional)
+  //            with api version
+  //   verb - action specific for different message kinds
+  //   args - action arguments
+  //
+  _createMessageWithArray(kind, target, verb, args) {
+    const message = {
+      [kind]: [this._nextMessageId, ...target],
+    };
+
+    if (verb) {
+      message[verb] = args;
+    }
+
+    this._nextMessageId += this._messageIdDelta;
+
+    return message;
   }
 
   // Create a JSTP message
@@ -394,7 +448,10 @@ class Connection extends EventEmitter {
     }
 
     const applicationName = message.handshake[1];
-    const application = this.server.applications[applicationName];
+    const applicationVersion = message.handshake[2];
+    const application = this.server._getApplication(
+      applicationName, applicationVersion
+    );
 
     if (!application) {
       this._handshakeError(errors.ERR_APP_NOT_FOUND);
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -147,10 +147,10 @@ class Connection extends EventEmitter {
     let message;
     if (version) {
       message = login && password ?
-        this._createPacketWithArray(
+        this._createMessageWithArray(
           'handshake', [name, version], 'login', [login, password]
         ) :
-        this._createPacketWithArray(
+        this._createMessageWithArray(
           'handshake', [name, version]
         );
     } else {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -128,7 +128,7 @@ class Connection extends EventEmitter {
   handshake(app, login, password, callback) {
     let name, version;
     if (typeof app === 'string') {
-      [name, version] = app.split('@');
+      [name, version] = common.rsplit(app, '@');
     } else {
       name = app.name;
       version = app.version;
--- a/lib/net.js
+++ b/lib/net.js
@@ -55,8 +55,8 @@ module.exports = {
   Server,
   createServer: (options, listener) =>
     createServer(options, listener),
-  connect: (appName, client, ...options) =>
-    connect(appName, client, ...options),
-  connectAndInspect: (appName, client, interfaces, ...options) =>
-    connectAndInspect(appName, client, interfaces, ...options),
+  connect: (app, client, ...options) =>
+    connect(app, client, ...options),
+  connectAndInspect: (app, client, interfaces, ...options) =>
+    connectAndInspect(app, client, interfaces, ...options),
 };
--- a/lib/server.js
+++ b/lib/server.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const semver = require('semver');
+
 const apps = require('./applications');
 const Connection = require('./connection');
 const SimpleAuthPolicy = require('./simple-auth-policy');
@@ -23,6 +25,16 @@ const initServer = function(
   if (Array.isArray(applications)) {
     applications = apps.createAppsIndex(applications);
   }
+
+  // versions cached for efficient search when provided version is a range
+  this._cachedVersions = new Map();
+  applications.forEach((appVersions, appName) => {
+    const versions = Array.from(appVersions)
+      .filter(version => version[0] !== 'latest')
+      .map(version => [new semver.SemVer(version[0]), version[1]])
+      .sort((a, b) => semver.rcompare(a[0], b[0]));
+    this._cachedVersions.set(appName, versions);
+  });
 
   if (typeof(authPolicy) === 'number') {
     heartbeatInterval = authPolicy;
@@ -82,6 +94,28 @@ class Server {
         connection.startHeartbeat(this.heartbeatInterval);
       }
     });
+  }
+
+  _getApplication(name, version) {
+    const appVersions = this.applications.get(name);
+    if (!appVersions) return null;
+    if (!version) return appVersions.get('latest');
+    // when version is not a range simply return matched
+    if (semver.valid(version)) return appVersions.get(version);
+
+    // search matching version, first matched will be the latest
+    try {
+      const range = new semver.Range(version);
+      const versions = this._cachedVersions.get(name);
+      for (let i = 0; i < versions.length; i++) {
+        // version === [versionCode, app]
+        const version = versions[i];
+        if (range.test(version[0])) return version[1];
+      }
+    } catch (error) {
+      // ignored
+    }
+    return null;
   }
 
   // Client connection event handler.
--- a/lib/simple-connect-policy.js
+++ b/lib/simple-connect-policy.js
@@ -14,16 +14,18 @@ module.exports = class SimpleConnectPolicy {
   // Should send handshake message with appropriate credentials
   // You can get client object provided upon connection creation
   // with connection.client.
+  //   app - string or object, application to connect to as 'name' or
+  //         'name@version' or { name, version }, where version
+  //         must be a valid semver range
   //   connection - JSTP connection
   //   callback - callback function that has signature
   //              (error, connection)
   //
-  connect(appName, connection, callback) {
+  connect(app, connection, callback) {
     connection.handshake(
-      appName, this.login, this.password,
+      app, this.login, this.password,
       (error) => {
         callback(error, connection);
-      }
-    );
+      });
   }
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1672,7 +1672,7 @@
         "pluralize": "4.0.0",
         "progress": "2.0.0",
         "require-uncached": "1.0.3",
-        "semver": "5.3.0",
+        "semver": "5.4.1",
         "strip-json-comments": "2.0.1",
         "table": "4.0.1",
         "text-table": "0.2.0"
@@ -4155,7 +4155,7 @@
       "requires": {
         "hosted-git-info": "2.5.0",
         "is-builtin-module": "1.0.0",
-        "semver": "5.3.0",
+        "semver": "5.4.1",
         "validate-npm-package-license": "3.0.1"
       }
     },
@@ -6632,9 +6632,9 @@
       "dev": true
     },
     "semver": {
-      "version": "5.3.0",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-5.3.0.tgz",
-      "integrity": "sha1-myzl094C0XxgEq0yaqa00M9U+U8="
+      "version": "5.4.1",
+      "resolved": "https://registry.npmjs.org/semver/-/semver-5.4.1.tgz",
+      "integrity": "sha512-WfG/X9+oATh81XtllIo/I8gOiY9EXRdv1cQdyykeXK17YcUW3EXUAi2To4pcH6nZtJPr7ZOpM5OMyWJZm+8Rsg=="
     },
     "set-blocking": {
       "version": "2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1158,7 +1158,7 @@
     "color-support": {
       "version": "1.1.3",
       "resolved": "https://registry.npmjs.org/color-support/-/color-support-1.1.3.tgz",
-      "integrity": "sha512-qiBjkpbMLO/HL68y+lh4q0/O1MZFj2RX6X/KmMa3+gJD3z+WwI1ZzDHysvqHGS3mP6mznPckpXmw1nI9cJjyRg==",
+      "integrity": "sha1-k4NDeaHMmgxh+C9S8NBDIiUb1aI=",
       "dev": true
     },
     "combined-stream": {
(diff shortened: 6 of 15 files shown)


lib/applications.js
```diff
@@ -56,10 +58,39 @@ apps.Application = Application;
...
 
   applications.forEach((application) => {
-    index[application.name] = application;
+    let appVersions = index.get(application.name);
+    if (!appVersions) {
+      appVersions = new Map();
+      index.set(application.name, appVersions);
+    }
+    if (!application.version) {
+      // no version means default version
```
nechaido: I'd change `default` to `latest`, but it is not necessary.

nechaido:
It should be mentioned that this PR Fixes: #31.

lib/applications.js
```diff
@@ -56,10 +58,39 @@ apps.Application = Application;
...
+      if (appVersions.get('latest')) {
```
belochub: You should definitely use `has` method here because you don't use the value.

lib/applications.js
```diff
@@ -56,10 +58,39 @@ apps.Application = Application;
...
+        throw new Error(
+          `Multiple entries of '${application.name} without version`
+        );
+      } else {
+        appVersions.set('latest', application);
+      }
+    } else {
+      appVersions.set(application.version.toString(), application);
```
belochub: Can version be a non-string value?
I thought that we agreed on using the semver, which meant using strings as version values.

lib/applications.js
```diff
@@ -56,10 +58,39 @@ apps.Application = Application;
...
+
+      // save latest version to fill missing latest versions later
+      const latestApp = latestApps.get(application.name);
+      if (!latestApp || application.version > latestApp.version) {
```
belochub: Same here, it is incorrect to compare semver versions like that.

lib/connection.js
```diff
@@ -112,15 +112,24 @@ class Connection extends EventEmitter {
...
   //   login - user name (optional)
   //   password - user password (optional)
   //   callback - callback function to invoke after the handshake is completed
   //
-  handshake(appName, login, password, callback) {
+  handshake(app, login, password, callback) {
+    let target;
+    if (typeof app === 'string') {
+      target = app;
+    } else if (!app.version) {
```
belochub: You should also add a possibility to provide `app` as a string containing both name and version like that: `appName@1.0.0`.

lib/applications.js
lundibundi: Sure.

lib/applications.js
```diff
@@ -9,11 +9,13 @@ module.exports = apps;
 // your needs.
 //   name - application name
 //   api - application API
+//   version - api version of 'name' application
```
lundibundi: Sure.

lib/applications.js
lundibundi: Oh, thanks.

lundibundi:
@belochub as discussed I've added support for semver in application versions. Also I've updated tests to check that functionality. Btw I've changed that comment message (api parameter description) event though this PR doesn't directly modify API but it is related to the API, so I think it's okay to put those changes here.
@metarhia/jstp-core PTAL.

lib/server.js
```diff
@@ -24,6 +26,15 @@ const initServer = function(
     applications = apps.createAppsIndex(applications);
   }
 
+  // versions cached for efficient search when provided version is a range
+  this._cachedVersions = {};
```
belochub: In my opinion, it is better to use `Object.create(null)` or even `Map` here.

lundibundi: Oh, I again forgot about map, thanks.

lib/server.js
```diff
@@ -84,6 +95,28 @@ class Server {
     });
   }
 
+  _getApplication(name, version) {
+    const appVersions = this.applications.get(name);
+    if (!appVersions) return null;
+    if (version) {
```
belochub: You can invert the condition here to avoid a lot of indentation below.

lib/connection.js
```diff
@@ -209,16 +221,21 @@ class Connection extends EventEmitter {
...
   //   args - action arguments
   //
   _createPacket(kind, target, verb, args) {
-    const packet = {
-      [kind]: (
-        target ? [this._nextPacketId, target] : [this._nextPacketId]
-      )
-    };
+    const packet = {};
+    if (Array.isArray(target)) {
```
aqrln: I don't like the idea of making this function polymorphic. It is a hot function, and passing arguments of different types will mess up with V8's ICs.

test/node/api-versions.js
```diff
@@ -0,0 +1,177 @@
...
+
+const interfacesV2 = {
+  calculator: {
+    answer(connection, callback) {
+      callback(null, 24);
+    },
+  }
+};
+
+const interfacesVLatest = {
```
aqrln: Wait... do I get it right that `latest` is a version by itself and not just a pointer to the latest semver version (`2.0.0` in this case)? That sounds wrong to me. `latest` and the like are tags, they must not be independent versions.

lib/connection.js
lundibundi: @aqrln I guess I'll split this into 2 functions with one accepting an array and the other value, what do you think?

test/node/api-versions.js
lundibundi: No, It was made to check specific behavior (when app has no version it's latest version). If every app has a version then the biggest version will be the latest, see https://github.com/metarhia/jstp/pull/231/files/66c24103eef081bdb56c42696a400182502634f5#diff-13dd866a30aedfb1fab7c3d9da9446ceR162.

belochub:
@aqrln, ping.

lib/connection.js
```diff
@@ -112,15 +113,26 @@ class Connection extends EventEmitter {
   }
 
   // Send a handshake packet over the connection
-  //   appName - name of an application to connect to
+  //   app - string or object, application to connect to as 'name' or
+  //         'name@version' or { name, version }
```
aqrln: Do we really need all the three of these? No strong opinion on it, but I'd rather make `name@version` the only format.

lib/connection.js
```diff
@@ -112,15 +113,26 @@ class Connection extends EventEmitter {
...
   //   password - user password (optional)
   //   callback - callback function to invoke after the handshake is completed
   //
-  handshake(appName, login, password, callback) {
+  handshake(app, login, password, callback) {
+    let target;
+    if (typeof app === 'string') {
+      const [name, version] = app.split('@');
+      if (semver.valid(version)) target = [name, version];
+      else target = name;
```
aqrln: I'd throw an error here instead of silently discarding an invalid version.

lib/connection.js
```diff
@@ -112,15 +113,26 @@ class Connection extends EventEmitter {
...
+    } else if (!app.version) {
+      target = app.name;
+    } else {
+      target = [app.name, app.version];
```
aqrln: Can you please stick to either a string (as in `{ handshake: [0, 'app@latest'] }`) or an array (as in `{ handshake: [0, 'app', 'latest'] }`) in both cases? As the parsed messages are plain objects, the current approach may suffer from the constantly changing object shapes.

lib/connection.js
aqrln: Yeah, that would work if we needed it at all. See my [other comment](https://github.com/metarhia/jstp/pull/231/files#r126623600).

lib/server.js
```diff
@@ -24,6 +26,16 @@ const initServer = function(
     applications = apps.createAppsIndex(applications);
   }
 
+  // versions cached for efficient search when provided version is a range
+  this._cachedVersions = new Map();
+  applications.forEach((appVersions, appName) => {
+    const versions = Array.from(appVersions)
+      .filter(app => app[0] !== 'latest')
+      .map(version => [new semver.SemVer(version[0]), version[1]])
```
aqrln: Can you please use destructuring in the left-hand side of the lambda to give some descriptive names for these values (i.e., `([something, somethingElse]) => [new semver.SemVer(something), somethingElse]`)?

Aside from that, why do `app` and `version` identifiers refer to the same objects in this and previous lines?

lib/connection.js
belochub: @aqrln, as far as I understand, first option you provided isn't possible, because it is always an array, check out `_createPacket` function. Sometimes version can just be omitted in the array, meaning that latest version should be used, and packet will look like this:
`{ handshake: [0, 'app'] }`.

lib/server.js
```diff
@@ -84,6 +96,29 @@ class Server {
...
+    try {
+      const range = new semver.Range(version);
+      const versions = this._cachedVersions.get(name);
+      for (let i = 0; i < versions.length; i++) {
+        // version === [versionCode, app]
+        const version = versions[i];
+        if (range.test(version[0])) return version[1];
+      }
+    } catch (error) {
+      // TODO (lundibundi): after loggers are implemented write using them
```
aqrln: Should it return a new error code instead of `ERR_APP_NOT_FOUND` in this case? 🤔

lib/server.js
belochub: @aqrln, there's comment on line 111 in this file.

lib/connection.js
aqrln: Ah, I see. Right.

lib/connection.js
aqrln: Why not always expect an array here though?

lib/server.js
aqrln: It doesn't help much. The file is read top to bottom, and these variable names are somewhat misleading. Both `app` and `version` at lines 33 and 34 are actually arrays of `[version, app]`, AFAIU.

belochub: Yes, but here it is obvious, because of line 32: this is an array created from the map with versions as keys, and applications as values, there is no need to make a destruction assignment here.
However, I do agree that arguments in lambdas should be renamed.

lib/connection.js
belochub: @aqrln, to avoid creation of an array with one element in some cases, it should probably be better for performance.

lib/connection.js
lundibundi: Same here, I don't really care. @metarhia/jstp-core somebody wants to leave the third option here?

lib/server.js
lundibundi: @belochub @aqrln Also I see no particular reason to specifically emphasize arguments with destructuring here and it should be sufficient to rename `.filter` argument to `version` to be consistent, bit I don't really care so it's up to you.

lib/server.js
lundibundi: @aqrln technically it's still missing application :smiley:, but it may indeed be clearer to actually add another error.

lib/connection.js
belochub: Yeah, why not?

lib/connection.js
lundibundi: Also I'd add that in most cases it is a single argument, as handshake is the only place with array and it's performed only once per connection.

lib/connection.js
lundibundi: No more comments means 1 for, 0 against, hence we left it as is.

belochub:
@lundibundi, can you address the comments and rebase this on master, please?

lundibundi:
@belochub I've already answered the comments and haven't seen any response since then, that's why I'm waiting.

belochub:
@lundibundi, you've agreed to rename the arguments (https://github.com/metarhia/jstp/pull/231#discussion_r126699371) and you haven't done it.
Also you haven't answered this https://github.com/metarhia/jstp/pull/231#discussion_r126623048 at all.

lundibundi:
@belochub  https://github.com/metarhia/jstp/pull/231#discussion_r126699371 - yes but I thought we should get the approval of @aqrln as he was the one who suggested to change it and not in the way we decided to. As for the https://github.com/metarhia/jstp/pull/231#discussion_r126623048 - you've answered  on that one, there is nothing to add.

belochub:
I haven't answered https://github.com/metarhia/jstp/pull/231#discussion_r126623048, you've probably confused it with some other comment.

lundibundi:
@belochub oh yeah, I confused it with the one below, thanks.

lib/connection.js
lundibundi: @aqrln Yeah, that will be better.

nechaido:
@aqrln ping.

nechaido:
@lundibundi ping see https://github.com/metarhia/jstp/pull/231#issuecomment-315419755.

lundibundi:
@nechaido As you can see in the comment right below the one you referenced I've already answered on the comments and waiting for the decision. As for the rebase, as I said before I want to finish it all at once (after addressing comments) and see no particular reason to rebase now.

nechaido:
@lundibundi as far as I can see neither https://github.com/metarhia/jstp/pull/231#discussion_r126623048 nor https://github.com/metarhia/jstp/pull/231#discussion_r126627605 a resolved. Correct me I am wrong.

belochub:
@nechaido, @lundibundi is waiting for @aqrln to answer the comments you mentioned.

nechaido:
@belochub it just seems to be that @lundibundi already agreed to make requested changes.

belochub:
@nechaido, he just wants to rebase and upload all of the changes at once as he said in https://github.com/metarhia/jstp/pull/231#issuecomment-316675765.

lib/applications.js
```diff
@@ -56,10 +64,39 @@ apps.Application = Application;
...
+      if (!latestApp || semver.gt(application.version, latestApp.version)) {
+        latestApps.set(application.name, application);
+      }
+    }
+  });
+
+  // set latest versions of apps without explicit latest version
+  latestApps.forEach((application, appName) => {
+    const appVersions = index.get(appName);
+    if (!appVersions.get('latest')) appVersions.set('latest', application);
```
aqrln: Nit: I'd prefer it to be multi-line.

lib/connection.js
aqrln: @lundibundi this is still unaddressed.

lib/server.js
aqrln: This is still unaddressed.

package-lock.json
```diff
@@ -713,7 +713,7 @@
     "color-support": {
```
aqrln: Can you please revert the `package-lock.json` changes, rebase onto the current `master` and reinstall `semver` with the latest npm?

lib/server.js
lundibundi: @aqrln but there is no approval of your to my proposed changes?

lib/server.js
lundibundi: @aqrln there is no response here as well.

lib/server.js
aqrln: @lundibundi I would still prefer `([version, app]) => ...`, but if you and @belochub are against it, then at least making the names consistent works for me too.

lib/connection.js
lundibundi: @aqrln as you can see in the comments below I'm waiting to get your approval to make all changes at once as I don't want to perform rebase multiple times for naught.

lib/server.js
aqrln: @lundibundi there haven't been any changes requested as well ;)

I'm fine with `ERR_APP_NOT_FOUND`.

lib/connection.js
lundibundi: @aqrln also what about this one?

aqrln: @lundibundi based on the conversation above, I though there was consensus about it. I have approved splitting this into two functions above, and neither me nor anyone else has objected to your last comment, so feel free to pick either approach.

lib/applications.js
```diff
@@ -54,10 +65,41 @@ class Application {
...
+      if (!latestApp || semver.gt(application.version, latestApp.version)) {
+        latestApps.set(application.name, application);
+      }
+    }
+  });
+
+  // set latest versions of apps without explicit latest version
+  latestApps.forEach((application, appName) => {
+    const appVersions = index.get(appName);
+    if (!appVersions.get('latest')) {
```
belochub: You should use `has` method here instead of `get` because you don't need the value, just checking if it exists.

lib/connection.js
```diff
@@ -109,15 +110,38 @@ class Connection extends EventEmitter {
...
   //
-  handshake(appName, login, password, callback) {
-    const packet = login && password ?
-      this._createPacket('handshake', appName, 'login', [login, password]) :
-      this._createPacket('handshake', appName);
+  handshake(app, login, password, callback) {
+    let name, version;
+    if (typeof app === 'string') {
+      [name, version] = app.split('@');
+      if (version && !semver.valid(version)) {
```
belochub: Shouldn't this method allow providing a range instead of exact version, if you enabled support for the ranges on the server side (`_getApplication` method in `lib/server.js`)?

lib/simple-connect-policy.js
```diff
@@ -6,9 +6,10 @@
 // anonymous handshake.
 //
 module.exports = class SimpleConnectPolicy {
-  constructor(login, password) {
+  constructor(login, password, version) {
```
belochub: Why did you choose to provide version here instead of passing it to the `connect` method?

aqrln: Good catch.

lundibundi: Hm, I thought in the simplest case user doesn't need to change the version dynamically and if they do, they will implement their own connect policy with such behavior. But yeah, I forgot about appName in `connect`, so I'll add version there.

lib/connection.js
```diff
@@ -120,8 +121,14 @@ class Connection extends EventEmitter {
     let name, version;
     if (typeof app === 'string') {
       [name, version] = app.split('@');
-      if (version && !semver.valid(version)) {
-        throw new Error('Invalid semver version');
+      if (version && !semver.validRange(version)) {
+        const error =  new Error('Invalid semver version range');
```
belochub: Redundant space here.

lib/connection.js
```diff
@@ -120,8 +121,14 @@ class Connection extends EventEmitter {
     let name, version;
     if (typeof app === 'string') {
       [name, version] = app.split('@');
-      if (version && !semver.valid(version)) {
-        throw new Error('Invalid semver version');
+      if (version && !semver.validRange(version)) {
```
belochub: Also, you should add a check for the version provided inside of an object, now you just assign its value even if it's incorrect (on line 135).

lib/simple-connect-policy.js
```diff
@@ -6,21 +6,21 @@
...
     this.login = login;
     this.password = password;
-    this.version = version;
   }
 
   // Should send handshake packet with appropriate credentials
   // You can get client object provided upon connection creation
   // with connection.client.
+  //   app - string or object, application to connect to as 'name' or
+  //         'name@version' or { name, version }
```
belochub: You should specify the constraints for `version` like you did in `lib/connection.js` on lines 114-115.

test/node/api-versions.js
```diff
@@ -74,10 +73,7 @@ test.test('must handle specific versions', (test) => {
     });
 
     // v1
-    client = {
-      connectPolicy: new jstp.SimpleConnectPolicy(null, null, '1'),
-    };
-    jstp.net.connect(app.name, client, port, (error, conn) => {
+    jstp.net.connect(app.name + '@1', null, port, (error, conn) => {
```
belochub: You know, that's the case for which passing app argument as an object, that I wanted to stay in this PR in https://github.com/metarhia/jstp/pull/231#discussion_r126700322, would come in handy.
This also goes for all of the similar occurrences in the tests.

lib/connection.js
```diff
@@ -364,7 +417,9 @@ class Connection extends EventEmitter {
     }
 
     const applicationName = packet.handshake[1];
-    const application = this.server.applications[applicationName];
+    const applicationVersion = packet.handshake[2];
+    const application =
```
belochub: I am not sure about the style of this, didn't we change styling for such cases?

lib/net.js
```diff
@@ -55,8 +55,8 @@ module.exports = {
   Server,
   createServer: (options, listener) =>
     createServer(options, listener),
-  connect: (appName, client, ...options) =>
-    connect(appName, client, ...options),
-  connectAndInspect: (appName, client, interfaces, ...options) =>
-    connectAndInspect(appName, client, interfaces, ...options),
+  connect: (app, client, ...options) =>
+    connect(app, client, ...options),
+  connectAndInspect: (app, client, interfaces, ...options) =>
```
belochub: You should also rename this argument in `lib/cli/command-processor.js`.

lib/applications.js
```diff
@@ -1,15 +1,26 @@
...
+//   api - object that contains interfaces as its fields each of which
+//         contains functions by method names. Each method has the following
+//         signature (connection, <0 or more method arguments>, callback)
+//   version - application version of 'name' application
 //
 class Application {
-  constructor(name, api) {
-    this.name = name;
+  constructor(name, api, version) {
+    [this.name, this.version] = name.split('@');
```
aqrln: Should it split from the right instead of from the left?

aqrln: I mean, `@metarhia/application@1.0` is a valid package name, so I'd expect it to be parsed as an application name correctly too.

lib/connection.js
```diff
@@ -109,15 +110,46 @@ class Connection extends EventEmitter {
...
   //   callback - callback function to invoke after the handshake is completed
   //
-  handshake(appName, login, password, callback) {
-    const packet = login && password ?
-      this._createPacket('handshake', appName, 'login', [login, password]) :
-      this._createPacket('handshake', appName);
+  handshake(app, login, password, callback) {
+    let name, version;
+    if (typeof app === 'string') {
+      [name, version] = app.split('@');
```
aqrln: Ditto.

lib/applications.js
belochub: @aqrln, in such cases you can provide an object with fields `name` and `version` set to corresponding values.

aqrln: @belochub I don't find it a valid justification. Why provide two ways to do the same thing with one of them being subtly broken? Either the npm-ish format must work according to users' expectations, or the `{ name, version }` schema must be the only available contract.

lundibundi: @aqrln and in the above example will the user have application name as `@metarhia/application`? As we don't support scopes (they mean nothing in our implementation) is there any reason to support that? But anyway it's easy to implement so it's okay with me.

belochub: @aqrln, by the way, are you sure that `@metarhia/application@1.0` is a valid package name?
Trying to create the package with such name using `npm` leads to the error:
`Sorry, name can only contain URL-friendly characters.`

aqrln: Did you try to make `@1.0` a part of the package name or...?

lib/applications.js
```diff
@@ -1,15 +1,27 @@
...
+//          'version' parameter
+//   api - object that contains interfaces as its fields each of which
+//         contains functions by method names. Each method has the following
+//         signature (connection, <0 or more method arguments>, callback)
+//   version - application version of 'name' application
 //
 class Application {
-  constructor(name, api) {
-    this.name = name;
+  constructor(name, api, version) {
```
nechaido: I'd prefer to swap parameters `api` and `version` but it's up to you.

test/node/api-versions.js
```diff
@@ -0,0 +1,206 @@
...
+  const serverConfig = {
+    applications: [appV1, appV2, appVlatest], authPolicy: app.authCallback,
+  };
+  server = jstp.net.createServer(serverConfig);
+  server.listen(0, () => {
+    test.plan(9);
+
+    const port = server.address().port;
+
+    // latest
```
nechaido: I'd prefer this to be a subtest.

test/node/api-versions.js
```diff
@@ -0,0 +1,206 @@
...
+    jstp.net.connect(appLatest, null, port, (error, conn) => {
+      test.assertNot(error, 'connect must not return an error');
+      conn.callMethod('calculator', 'answer', [], (error, result) => {
+        test.assertNot(error, 'callMethod must not return an error');
+        test.strictSame(result, 13);
+        conn.close();
+      });
+    });
+
+    // v1
```
nechaido: Ditto.

test/node/api-versions.js
```diff
@@ -0,0 +1,206 @@
...
+    jstp.net.connect(appV1, null, port, (error, conn) => {
+      test.assertNot(error, 'connect must not return an error');
+      conn.callMethod('calculator', 'answer', [], (error, result) => {
+        test.assertNot(error, 'callMethod must not return an error');
+        test.strictSame(result, 42);
+        conn.close();
+      });
+    });
+
+    // v2
```
nechaido: Ditto.

test/node/api-versions.js
```diff
@@ -0,0 +1,206 @@
...
+  const serverConfig = {
+    applications: [appV1, appV2, appVlatest], authPolicy: app.authCallback,
+  };
+  server = jstp.net.createServer(serverConfig);
+  server.listen(0, () => {
+    test.plan(6);
+
+    const port = server.address().port;
+
+    // ^1.0.0
```
nechaido: Ditto.

test/node/api-versions.js
```diff
@@ -0,0 +1,206 @@
...
+    jstp.net.connect(appV1Compatible, null, port, (error, conn) => {
+      test.assertNot(error, 'connect must not return an error');
+      conn.callMethod('calculator', 'answer', [], (error, result) => {
+        test.assertNot(error, 'callMethod must not return an error');
+        test.strictSame(result, 42);
+        conn.close();
+      });
+    });
+
+    // > v1.0.0
```
nechaido: Ditto.

lib/applications.js
lundibundi: Even though I agree that it will look better first of all it will not be compatible and second in case version is specified in the `name` you will have to pass `null` to the version parameter, because api will come after it.

belochub:
Landed in https://github.com/metarhia/jstp/commit/9116125e6dce8ab5a99519790693f37a5660937e.

PR #245, opened 2017-06-28
lib: rewrite i-face introspection without promises

tshemsedinov:


lib/transport-common.js
```diff
@@ -68,25 +68,23 @@ const newConnectAndInspectFn = (connFactory, transportClass) => {
...
-        connection.inspectInterface(name, (error, proxy) => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve(proxy);
+      const proxies = {};
+      let errored = false;
+      const len = interfaces.length;
+      let cnt = 0;
+      interfaces.map(name => connection.inspectInterface(name,
```
belochub: What's the point of using a `map` method here, if the result is being ignored in the end?
Maybe replace it with `forEach`?

lib/transport-common.js
```diff
@@ -68,25 +68,23 @@ const newConnectAndInspectFn = (connFactory, transportClass) => {
...
     connect(appName, client, ...options, (error, connection) => {
       if (error) return callback(error);
-
-      Promise.all(interfaces.map(name => new Promise((resolve, reject) => {
-        connection.inspectInterface(name, (error, proxy) => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve(proxy);
+      const proxies = {};
```
belochub: This object should be created with `Object.create(null)` like it was done before this patch.

tshemsedinov: @belochub yes, it's better, thanks

lib/transport-common.js
tshemsedinov: @belochub sure, it's better to use even `for` loop here

belochub:
Name and commit message of this PR exceed the 50 character limit, it should be renamed.

lib/transport-common.js
```diff
@@ -68,25 +68,23 @@ const newConnectAndInspectFn = (connFactory, transportClass) => {
...
-      Promise.all(interfaces.map(name => new Promise((resolve, reject) => {
-        connection.inspectInterface(name, (error, proxy) => {
-          if (error) {
-            reject(error);
-          } else {
-            resolve(proxy);
+      const proxies = Object.create(null);
+      let errored = false;
+      const len = interfaces.length;
+      let cnt = 0;
```
belochub: In my opinion, it is better to rename this variable to full name (`count`) to avoid possible misunderstanding.

tshemsedinov:
@belochub agree, fixed, who will merge?

belochub:
Landed in https://github.com/metarhia/jstp/commit/43c61e7e454dda7d0b867c376c598ad68c52745c.

PR #246, opened 2017-06-28
lib: prepare code for metarhia-common and metasync

tshemsedinov:
@aqrln This PR is not complete, just for your interest. I'll do it in multiple steps and I may need your help.


tshemsedinov:
@aqrln done, but I rebased this branch with branch `style-and-opt` to avoid conflicts with eb7a704 `lib: refactor return style` so PR includes commit from different PR.

tshemsedinov:
@aqrln rebased again

lib/remote-proxy.js
```diff
@@ -55,15 +57,8 @@ RemoteProxy.wrapRemoteMethod = (instance, methodName) => {
...
-  let callback = args[args.length - 1];
-  args = Array.prototype.slice.call(args, 0, -1);
-
-  if (typeof(callback) !== 'function') {
-    args.push(callback);
-    callback = null;
-  }
-
-  this._connection.callMethod(this._interfaceName, methodName, args, callback);
+  const cb = common.cbExtract(args);
```
aqrln: Can we please call it `callback`? I'd prefer to use descriptive names when possible.

lib/socket.js
```diff
@@ -83,12 +85,12 @@ class Transport extends EventEmitter {
 }
 
 const socketFactory = connect => (...options) => {
-  const callback = common.extractCallback(options);
+  const cb = common.cbExtract(options);
```
aqrln: Ditto.

lib/transport-common.js
```diff
@@ -27,10 +29,10 @@ const common = require('./common');
 const newConnectFn = (
   connFactory, transportClass
 ) => (appName, client, ...options) => {
-  const callback = common.extractCallback(options);
+  const cb = common.cbExtract(options);
```
aqrln: Ditto.

lib/transport-common.js
```diff
@@ -71,30 +73,21 @@ const newConnectFn = (
 const newConnectAndInspectFn = (connFactory, transportClass) => {
   const connect = newConnectFn(connFactory, transportClass);
   return (appName, client, interfaces, ...options) => {
-    const callback = common.extractCallback(options);
+    const cb = common.cbExtract(options);
```
aqrln: Ditto.

lib/transport-common.js
```diff
@@ -71,30 +73,21 @@ const newConnectFn = (
...
       if (error) {
-        callback(error);
+        cb(error);
         return;
       }
-      const proxies = Object.create(null);
-      let errored = false;
-      const len = interfaces.length;
-      let count = 0;
+      const pCollector = metasync.collect(interfaces.length);
```
aqrln: Maybe something like `proxiesCollector` or even just `collector`. That lone "p" letter doesn't help readability much and looks like hungarian notation.

lib/ws-browser.js
```diff
@@ -80,14 +82,15 @@ class Transport extends EventEmitter {
 //   appProvider - client application provider
 //
 const socketFactory = (url, callback) => {
+  const cb = common.cb(callback);
```
aqrln: I'm not sure whether we need it here, it may be handled at a higher level.

/cc @lundibundi

lib/ws-internal.js
```diff
@@ -124,13 +126,13 @@ const initServer = function(options, httpServer) {
 //             is established.
 //
 const connectionFactory = (webSocketConfig, ...options) => {
-  const callback = common.extractCallback(options);
+  const cb = common.cbExtract(options);
```
aqrln: I'd prefer it to be `callback`.

package-lock.json
```diff
@@ -45,12 +45,14 @@
       }
     },
     "ajv": {
-      "version": "4.11.8",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.8.tgz",
-      "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
+      "version": "5.2.2",
```
aqrln: There's a huge deal of unrelated changes and version bumps. Can you please revert all the changes to `package.json` and `package-lock.json` and run `npm i metasync metarhia-common` anew?

tools/common.js
```diff
@@ -24,4 +13,18 @@ common.promisify = fn => (...args) => (
...
+  const exec = promisify(childProcess.exec);
+  return exec(cmd).then((stdout, stderr) => {
+    if (stderr) console.error(stderr);
+    return stdout;
+  });
+};
+
+const writeFile = promisify(fs.writeFile);
+
+module.exports = {
```
aqrln: This change is unrelated. Can you please split it into another PR?

lib/ws-browser.js
lundibundi: This is also an internal function, there is always a callback here.

lundibundi: This is also an internal function, there is always a callback here.

package-lock.json
tshemsedinov: @aqrln but this PR requires `metasync` and `metarhia-common` on `package.json`

aqrln: Yes. But running `npm install metasync metarhia-common` will add them to `package.json` too.

tshemsedinov:
@aqrln see dff1d345d2b4c341119919f2b7b75d1ccc6d5ec4 it is related to this PR because of naming but extracted to separate commit.

tshemsedinov:
@aqrln rebased and ready

lib/cli/line-processor.js
```diff
@@ -75,13 +75,13 @@ module.exports = class LineProcessor {
       callback(reportMissingArgument('Host'));
       return;
     }
-    const args = utils.split(tokens, ' ', 2);
-    let [scheme, authority] = utils.split(args[0], '://', 1, true);
+    const args = autocomplete.split(tokens, ' ', 2);
```
belochub: `split` function here has nothing to do with autocomplete whatsoever, it is just used to split up the arguments provided by the user in CLI.

package.json
```diff
@@ -37,6 +37,8 @@
   "dependencies": {
     "semver": "^5.4.1",
     "uuid": "^3.1.0",
+    "metarhia-common": "0.0.19",
```
nechaido: Shouldn't it be: `"metarhia-common": "^0.0.19",`?

lib/cli/autocomplete.js
```diff
@@ -94,3 +91,11 @@ utils.iterativeCompletion = (inputs, depth, completer) => {
   }
   return [completions, help];
 };
+
+module.exports = {
+  filterKeys,
+  split,
```
nechaido: Why have you moved `rsplit` to `metarhia-common` but have not moved `split`?

lib/socket.js
```diff
@@ -84,7 +85,7 @@ class Transport extends EventEmitter {
 
 const socketFactory = (connect) => {
   const resultConnectFn = (...options) => {
-    const callback = common.extractCallback(options);
+    const callback = common.cbExtract(options);
```
lundibundi: `extractCallback` and `cbExtract` are not equivalent, the latter has a lot of checks that are completely unnecessary here. I think it's better to place `extractCallback` in common and rename `cbExtract` to something that will represent its purpose, like `cbWrapExtract`.

lib/transport-common.js
```diff
@@ -29,7 +31,7 @@ const common = require('./common');
 const newConnectFn = (
   connFactory, transportClass
 ) => (app, client, ...options) => {
-  const callback = common.extractCallback(options);
+  const callback = common.cbExtract(options);
```
lundibundi: Same as [cbExtract comment](https://github.com/metarhia/jstp/pull/246/files#r131100198).

lib/transport-common.js
```diff
@@ -73,30 +75,21 @@ const newConnectFn = (
 const newConnectAndInspectFn = (connFactory, transportClass) => {
   const connect = newConnectFn(connFactory, transportClass);
   return (app, client, interfaces, ...options) => {
-    const callback = common.extractCallback(options);
+    const callback = common.cbExtract(options);
```
lundibundi: Same as [cbExtract comment](https://github.com/metarhia/jstp/pull/246/files#r131100198).

lib/ws-internal.js
```diff
@@ -124,7 +125,7 @@ const initServer = function(options, httpServer) {
 //             is established.
 //
 const connectionFactory = (webSocketConfig, ...options) => {
-  const callback = common.extractCallback(options);
+  const callback = common.cbExtract(options);
```
lundibundi: Same as [cbExtract comment](https://github.com/metarhia/jstp/pull/246/files#r131100198).

package.json
aqrln: @nechaido both `0.0.19` and `^0.0.19` mean the same thing. Caret ranges only allow updates that don't modify the leftmost non-zero part of the version.

lib/applications.js
```diff
@@ -17,7 +18,7 @@ const errors = require('./errors');
 //
 class Application {
   constructor(name, api, version) {
-    [this.name, this.version] = common.rsplit(name, '@');
+    [this.name, this.version] = common.rsection(name, '@');
```
lundibundi: I think that it's better to update `rsplit` function to support cases from [python](https://docs.python.org/3/library/stdtypes.html#str.rsplit) where I actually took it from and name it as such.

lib/socket.js
tshemsedinov: Agree, in `metarhia-common` v0.0.21 we have `safeCallback` (renamed `cbExtract`) and `unsafeCallback` (renamed `cbUnsafe`). I placed your code of `extractCallback` into `common.unsafeCallback` because it's optimal implementation. Also I renamed `common.cb` to `once` because `once` will be clear and native for js world.

lib/applications.js
tshemsedinov: I agree, we need full `rsplit` so I added issue here: https://github.com/metarhia/common/issues/91 but I think in this place `rsection` is ok because it is simpler and in multiple other cases `rsection` will be optimal.

lib/cli/autocomplete.js
tshemsedinov: I changed `rsplit` to `common.rsection` because it is not compatible with `String.prototype.split()` This `split` is also not compatible with `String.prototype.split()` so I think we will either rename it or use `String.prototype.split()` instead.

lib/cli/line-processor.js
tshemsedinov: I agree, but see https://github.com/metarhia/jstp/pull/246/files#r132857348 We must arrange such common utilities to have code consistency over project, Meterhia stack and JS stories in general.

lib/transport-common.js
tshemsedinov: https://github.com/metarhia/jstp/pull/246/files#r132848995

lib/transport-common.js
tshemsedinov: https://github.com/metarhia/jstp/pull/246/files#r132848995

lib/ws-internal.js
tshemsedinov: https://github.com/metarhia/jstp/pull/246/files#r132848995

lib/transport-common.js
tshemsedinov: In this case user can call it without callback so we need `common.safeCallback`

lib/socket.js
tshemsedinov: In this case user can call it without callback so we need `common.safeCallback`

lib/transport-common.js
tshemsedinov: In this case user can call it without callback so we need `common.safeCallback`

lib/ws-internal.js
tshemsedinov: In this case user can call it without callback so we need `common.safeCallback`

tshemsedinov:
In most cases we use `extractCallback` (new name `unnsafeCallback`) to get callback from user-defined connection function, so user can pass `callback` optionally. If no callback passed exception will throw. In such cases we may choose one of the following ways:
- Use `unsafeCallback` and declare callback as required, so user will be responsible for inappropriate use.
- Implement `requiredCallback` that requires last argument to be callback and throw exception otherwise.
- Use `safeCallback` and have no problems but overhead.
- Change `safeCallback` behavior: use empty function but remove prevention of multiple calls, so it will be safe enough for above this cases. For more strict cases we can implement `onceCallback`.

I'd prefer to have in `metarhia-common` all those cases:
- `unsafeCallback(array): function or null`
- `safeCallback(array): function or emptiness`
- `requiredCallback(array): function or throw`
- `onceCallback(array): once(function) or emptiness`

@metarhia/jstp-core

lib/socket.js
lundibundi: This is an internal function that always has callback, the user cannot call it.

lib/transport-common.js
lundibundi: In this case not providing a callback is both an error and strange (you are calling a function to create a connection and not expecting a result), that's why I used extract here.

lib/transport-common.js
lundibundi: Same as https://github.com/metarhia/jstp/pull/246#discussion_r132892967.

lib/ws-internal.js
lundibundi: Same as https://github.com/metarhia/jstp/pull/246#discussion_r132892234.

lib/socket.js
tshemsedinov: This callback comes from `jstp.net.connect` and `jstp.net.connectAndInspect` and tls/ws/wss analogs. Try to remove this callack in `test/node/transport-net-tcp.js` for example.

nechaido: At the time it throws a TypeError:
```
> jstp.net.connect('name', null, 123);
undefined
> TypeError: callback is not a function
    at connFactory (/home/nechaido/Projects/metarhia/jstp/lib/transport-common.js:35:7)
    at Socket.onceErrorListener (/home/nechaido/Projects/metarhia/jstp/lib/socket.js:91:9)
    at Object.onceWrapper (events.js:316:30)
    at emitOne (events.js:115:13)
    at Socket.emit (events.js:210:7)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```
I am ok with it. `connect`is not a function that can be called without callback.

tshemsedinov: We need to avoid throwing, it terminates process, it's better to bubble an error to upper level without process termination.

tshemsedinov: But I can agree that `connect` is client-side function, client-side code is less durable in termination/restarting meaning.

belochub: @tshemsedinov, but it's a completely natural behavior in Node.js for `TypeError` to occur whenever you provide invalid arguments to the function (which is the case with providing no callback to the `connect` function).
This means that the user gets `TypeError` thrown only when he/she is using the library incorrectly.

belochub:
Closing this PR, since it's become outdated to the point where it's easier to rewrite it from scratch compared to resolving all of the merge conflicts when rebasing on `master`.
I will also create a corresponding issue to keep as a reminder (as proposed by @tshemsedinov).

Issue #139, opened 2017-05-06
Support connecting via Unix domain sockets / named pipes in CLI


aqrln:
Please see https://github.com/metarhia/JSTP/pull/138#pullrequestreview-36646313 and https://github.com/metarhia/JSTP/pull/138#issuecomment-299653465 for the background.

lundibundi:
@belochub ping, will you add this one or should I do it?

PR #251, opened 2017-07-10
cli: support connections via Unix domain sockets

belochub:
Also support connections via named pipes on Windows systems.

Fixes: https://github.com/metarhia/jstp/issues/139

### DIFF
--- a/lib/cli/line-processor.js
+++ b/lib/cli/line-processor.js
@@ -66,14 +66,18 @@ module.exports = class LineProcessor {
       scheme = DEFAULT_SCHEME;
     }
     const [host, portString] = utils.split(authority, ':', 2, true);
+    let port;
     if (!host) {
       return callback(reportMissingArgument('Host'));
-    } else if (scheme !== 'ipc' && !portString) {
-      return callback(reportMissingArgument('Port'));
     }
-    const port = Number(portString);
-    if (scheme !== 'ipc' && (isNaN(port) || port < 0 || port >= 65536)) {
-      return callback(new Error(`Port has incorrect value: ${portString}`));
+    if (scheme !== 'ipc') {
+      if (!portString) {
+        return callback(reportMissingArgument('Port'));
+      }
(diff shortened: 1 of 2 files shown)


lib/cli/line-processor.js
```diff
@@ -68,11 +68,11 @@ module.exports = class LineProcessor {
     const [host, portString] = utils.split(authority, ':', 2, true);
     if (!host) {
       return callback(reportMissingArgument('Host'));
-    } else if (!portString) {
+    } else if (scheme !== 'ipc' && !portString) {
```
lundibundi: Maybe it's better to put everything under `!== 'ipc'` case instead of checking it 2 times, it may be cleaner in my opinion.

belochub: What is 'everything', can you clarify this bit?

lundibundi: @belochub lines 71-77.

belochub: @lundibundi, it is not possible, because then `port` will be declared inside this block (it is on line 74) and won't be visible as an argument to `commandProcessor.connect` function.
Can you explain your idea, please, because I don't understand how you want this to look like?

lundibundi: @belochub just declare port before `if (scheme === 'ipc')` check and then put lines 71-77 under that `if`.
```javascript
const port = 0;
if (scheme !== 'ipc') {
  if (!portString) {
    return callback(reportMissingArgument('Port'));
  }
  port = Number(portString);
  if (isNaN(port) || port < 0 || port >= 65536) {
    return callback(new Error(`Port has incorrect value: ${portString}`));
  }
}
```

belochub:
@lundibundi, is it okay now?

belochub:
Landed in https://github.com/metarhia/jstp/commit/2f2f75bdb753f127b079b5f391a6cb189ebd6611.

Issue #250, opened 2017-07-10
Emit events about network messages on connections


aqrln:
We used to have such events, but they weren't thoroughly planned, conveyed lots of garbage info, led to deopts, and, eventually, got removed in #222. That was a bit radical because we still need this functionality to move logging away from this package (this will make #224 basically unnecessary).

Refs: #49

PR #252, opened 2017-07-11
lib: emit events about connection messages

lundibundi:
As discussed I add events to the connection, also events about connection establishment and destruction already exist on server ('connect', 'disconnect'). 
Also I did not forward connection events through a server as this will most likely have a significant performance cost, especially during highload. And it's easy to implement if user ever needs it:
```javascript
server.on('connect', (connection) => {
  common.forwardMultipleEvents(connection, server,
    ['handshakeRequest', 'call', 'event', 'inspect']);
});
```
Fixes: https://github.com/metarhia/jstp/issues/250


lib/connection.js
```diff
@@ -384,6 +384,8 @@ class Connection extends EventEmitter {
 
     this.server.startSession(this, application, authStrategy, credentials,
       this._onSessionCreated.bind(this));
+
+    this.emit('handshakeRequest', applicationName, authStrategy, credentials);
```
belochub: I don't think we should emit credentials here.

belochub:
@lundibundi, we also discussed the possibility to log all of the incoming/outgoing messages in debug mode, and I haven't found the implementation of corresponding events here. Will you add it later?

lib/connection.js
lundibundi: @belochub also not sure about that, but I thought that users may decide for themselves whether they need to log them or just ignore this parameter. @metarhia/jstp-core?

lundibundi:
@belochub yeah, but should we just emit them as well when node is in debug mode?

lib/connection.js
belochub: @lundibundi, it won't be secure that way.

belochub:
@lundibundi, we discussed that there will be different modes and in "debug" or "verbose" mode all of the messages will be emitted.

lib/connection.js
lundibundi: @belochub It won't be secure it user decides to log them, but they may have different trade-offs there.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,186 @@
...
+const test = require('tap');
+const net = require('net');
+
+const jstp = require('../..');
+
+const app = require('../fixtures/application');
+
+const application = new jstp.Application(app.name, app.interfaces);
+const serverConfig =
+  { applications: [application], authPolicy: app.authCallback };
```
aqrln: Style nit: maybe split this object over several lines? It's a bit long.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,186 @@
...
+
+    let serverConnection = null;
+    server.once('connect', (serverConn) => {
+      serverConnection = serverConn;
+    });
+
+    const port = server.address().port;
+    const socket = net.connect(port);
+    socket.on('error',
+      () => test.fail('must create socket and connect to server'));
```
aqrln: Nit: can you please use braces here?

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,186 @@
...
+    socket.on('connect', () => {
+      serverConnection.on('handshakeRequest',
+      (applicationName, authStrategy, credentials) => {
```
aqrln: There's a missing indentation level.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,186 @@
...
+
+    let serverConnection = null;
+    server.once('connect', (serverConn) => {
+      serverConnection = serverConn;
+    });
+
+    const port = server.address().port;
+    const socket = net.connect(port);
+    socket.on('error',
+      () => test.fail('must create socket and connect to server'));
```
aqrln: Could you please add braces here too?

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,186 @@
...
+    socket.on('connect', () => {
+      serverConnection.on('handshakeRequest',
+      (applicationName, authStrategy, credentials) => {
```
aqrln: Missing indentation.

test/node/connection-emit-actions.js
lundibundi: Yeah, I forgot to update local eslint.

test/node/connection-emit-actions.js
lundibundi: Sure

test/node/connection-emit-actions.js
lundibundi: Sure. Btw we've got the same code in `connection-call.js`, We should probably also change it there.

lib/connection.js
belochub: @lundibundi, ping. You haven't changed it.

lundibundi: Yes because I see no need to, we are not responsible for users breaking their security by themselves.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,192 @@
...
+          'session id must be equal to the one provided by authCallback');
+      });
+
+      connection.handshake(app.name, null, null, (error) => {
+        test.assertNot(error, 'handshake must not return an error');
+        test.equal(connection.username, null, 'username must be null');
+        test.equal(connection.sessionId, app.sessionId,
+          'session id must be equal to the one provided by authCallback');
+        connection.close();
+        test.end();
```
nechaido: Missed this bit, is it absolute that this callback will be called after both `handshake` event handlers? If not - test will fail because of assertions after end, in that case the right way to do it is using `test.plan()`. Same applies to every test here.

lundibundi: Yes, that's why I made them like this. Though that's implementation specific to be precise, so maybe it'll be better to use plan? @metarhia/jstp-core should we switch to plan here?

nechaido: IMHO, if we multiple callbacks\handlers must be called, it is better to use `test.plan()`.

lundibundi: Oh well, nobody seems to bother, but I guess the less implementation specific test are the better so I'll change it.

lundibundi:
@metarhia/jstp-core ping.

lib/connection.js
```diff
@@ -377,6 +397,8 @@ class Connection extends EventEmitter {
       this, application, authStrategy, credentials,
       this._onSessionCreated.bind(this)
     );
+
+    this.emit('handshakeRequest', applicationName, authStrategy, credentials);
```
belochub: @lundibundi, I am still against emitting credentials here.

lundibundi: @belochub And I still see no harm in doing that. Let's wait what others have to say on this matter.

nechaido: @belochub what if we emit them only in debug mode, but in release mode we omit them?

lundibundi: It makes no difference, in debug mode you have whole packets logged, so there is no reason to make such a switch.

nechaido: @lundibundi if you put it this way, I agree with @belochub because I see no use in emiting credential in release mode.

belochub: @lundibundi, that's why I'm opting for not emitting them here: if a user will need to debug application including authentication he/she can easily do that by getting complete packets.
And outside of the debug mode emitting credentials makes no sense, no one should use them in that case, because logging them will lead to security problems (and you agree on this).

nechaido:
@tshemsedinov ping.

lib/connection.js
lundibundi: @belochub Okay, I'll remove it soon.

lib/connection.js
```diff
@@ -402,7 +402,7 @@ class Connection extends EventEmitter {
       this._onSessionCreated.bind(this)
     );
 
-    this.emit('handshakeRequest', applicationName, authStrategy, credentials);
```
belochub: You forgot to update the tests correspondingly.

lib/connection.js
```diff
@@ -193,6 +197,10 @@ class Connection extends EventEmitter {
   _heartbeatCallback(interval) {
     this.transport.send('{}');
     this.setTimeout(interval, this._heartbeatCallbackInstance);
+
+    if (process.env.NODE_ENV === 'debug') {
```
belochub: I am not so sure about the "debug" value, as far as I know, "production" and "development" are used as values for the `NODE_ENV` environment variable.

lib/connection.js
```diff
@@ -193,6 +197,10 @@ class Connection extends EventEmitter {
   _heartbeatCallback(interval) {
     this.transport.send('{}');
     this.setTimeout(interval, this._heartbeatCallbackInstance);
+
+    if (process.env.NODE_ENV !== 'production') {
+      this.emit('sentPacket', this._heartbeatMessage);
```
belochub: Shouldn't there be tests for this event?

lib/connection.js
```diff
@@ -307,6 +323,10 @@ class Connection extends EventEmitter {
   //   packets - array of packets
   //
   _processPacket(packet) {
+    if (process.env.NODE_ENV !== 'production') {
+      this.emit('receivedPacket', packet);
```
belochub: This event should probably be tested as well.

lib/connection.js
lundibundi: Good idea, I thought so too, but is it okay to change env variable here to test it?

belochub: You don't actually need to change it, it will "development" by default.

lundibundi: Yes, but it'll be strange if people will have test failing if they start them in console with NODE_ENV set to 'production', from maybe, the start of their server or anything.

lundibundi: Actually I think it'll be okay to test both behaviors based on current env variable value.

belochub: @lundibundi, well, tests are for development mode, not production mode.

lundibundi:
@metarhia/jstp-core if anyone has a better solution to check if packet is one of provided feel free to suggest (see `addEmitPacketCheck` and `test.oneOf`).

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+        test.equal(connection.username, null, 'username must be null');
+        test.equal(connection.sessionId, app.sessionId,
+          'session id must be equal to the one provided by authCallback');
+        connection.close();
+      });
+    });
+  }
+);
+
+test.test('must emit server and client events upon login/password handshake',
```
belochub: It is called "login authentication strategy".

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+    socket.on('error', () => {
+      test.fail('must create socket and connect to server');
+    });
+    socket.on('connect', () => {
+      serverConnection.on('handshakeRequest',
+        (applicationName, authStrategy) => {
+          test.equal(applicationName, app.name,
+            'application name must match');
+          test.equal(authStrategy, 'login',
+            'auth strategy must be \'login\' with login/password');
```
belochub: Rename "auth" into "authentication" for clarity ("auth" can also mean "authorization").

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+  (test) => {
+    test.plan(5);
+
+    const iface = 'calculator';
+    const methodName = 'doNothing';
+    const args = [];
+    addCallEventCheck(test, server.getClients()[0], iface, methodName, args);
+    connection.on('callback', (error, ok) => {
+      test.assertNot(error, 'callMethod must not return an error');
+      test.strictSame(ok, [], 'Ok contents must match');
```
belochub: You should decapitalize "Ok" in the message to keep it consistent.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+  (test) => {
+    test.plan(5);
+
+    const iface = 'calculator';
+    const methodName = 'answer';
+    const args = [];
+    addCallEventCheck(test, server.getClients()[0], iface, methodName, args);
+    connection.on('callback', (error, ok) => {
+      test.assertNot(error, 'callMethod must not return an error');
+      test.strictSame(ok, [42], 'Ok contents must match');
```
belochub: Ditto.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+    addCallEventCheck(test, server.getClients()[0], iface, methodName, args);
+    connection.on('callback', (error, ok) => {
+      test.assertNot(error, 'callMethod must not return an error');
+      test.strictSame(ok, [], 'Ok contents must match');
+    });
+    connection.callMethod(iface, methodName, args);
+  }
+);
+
+test.test('must emit event on call with no arguments and return value',
```
belochub: "with no arguments and return value" is a bit misleading: it is not clear if "no" does also relate to "return value".

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+    test.assert(expectedInterfaces.includes(interfaceName),
+      'inspect event interface must be one of expected');
+  });
+
+  expectedInterfaces.forEach((iface) => {
+    connection.inspectInterface(iface);
+  });
+});
+
+test.test('must emit packets in not "production" mode', (test) => {
```
belochub: It's either "development" or "production" mode, you can replace "not production" with "development".

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+  // 4 packets from call below and 4 from 1 heartbeat
+  test.plan(8);
+
+  const clientSentPackets = [{}, { call: [1, 'calculator'], answer: [] }];
```
nechaido: I'd prefer this test not to start handshake. this way there will be no need for `tap-oneOf`.

lib/connection.js
nechaido: Heartbeat should have it's own `heartbeat` event.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
+      serverConnection = serverConn;
+    });
+
+    const port = server.address().port;
+    const socket = net.connect(port);
+    socket.on('error', () => {
+      test.fail('must create socket and connect to server');
+    });
+    socket.on('connect', () => {
+      serverConnection.on('handshakeRequest',
```
nechaido: IMHO this test is not redundant.

lib/connection.js
lundibundi: @nechaido I see no reason for this it just a packet as everything else. @metarhia/jstp-core what do you think?

test/node/connection-emit-actions.js
lundibundi: @nechaido You meant heartbeat? I don't really care, I thought that we should test that it emits heartbeat packets.

lib/connection.js
lundibundi: Done.

test/node/connection-emit-actions.js
lundibundi: Thanks.

test/node/connection-emit-actions.js
lundibundi: Okay.

test/node/connection-emit-actions.js
lundibundi: Also we should change this in the connection tests, as it's the same there.

test/node/connection-emit-actions.js
lundibundi: I'd test both as I see no harm in doing so.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,221 @@
...
```
lundibundi: I just wanted to put them in one place, also this callback should always be called after the one above, but it's okay to put them there too.

test/node/connection-emit-actions.js
lundibundi: Okay

lib/connection.js
belochub: @lundibundi, I think, @nechaido meant that there should also be `heartbeat` event, which will be always emitted in both development and production modes.
However, I don't see think we need this, emitting this messages in development mode only seems enough.

test/node/connection-emit-actions.js
nechaido: @lundibundi I mean that we should test them separately.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,210 @@
...
+  });
+
+  connection.callMethod('calculator', 'answer', []);
+});
+
+test.test('must emit heartbeat messages in development mode', (test) => {
+  test.plan(2);
+
+  server.getClients()[0].on('heartbeat', (message) => {
+    test.strictSame(message, {}, 'Heartbeat message must match on server side');
```
belochub: You should decapitalize "Heartbeat" here.

test/node/connection-emit-actions.js
```diff
@@ -0,0 +1,210 @@
...
+  });
+  connection.on('heartbeat', (message) => {
+    test.strictSame(message, {}, 'Heartbeat message must match on client side');
```
belochub: Here as well.

belochub:
Landed in https://github.com/metarhia/jstp/commit/04ed72e8eebf5f0653fc36ca5798bd36f1328879.

PR #253, opened 2017-07-11
bench: add simple benchmark

nechaido:


lundibundi:
@metarhia/jstp-core After playing around with it for a bit I found out that our server doesn't handle highload at all and just drops with OOM, maybe should we file an issue and think for some kinds of measures against this?

belochub:
@lundibundi, it is impossible to overwrite user's data by listening on a socket which is an existing file, `listen` just won't permit the creation of the socket in that case.
Also, benchmark always removes socket file if it finishes correctly, and there is no way you can remove the file after crashing with OOM error, so it does not solve this problem. We can probably add a listener removing the file to SIGINT event, it will help in cases when the benchmark is stopped manually.

lundibundi:
@belochub Ok, removed that point, it's not necessary to find ways to remove the file if server crashed with OOM, although SIGINT should probably be handled.

nechaido:
@lundibundi added error and `SIGINT` handling.
`Could not connect to the server: Error: connect EAGAIN /tmp/jstp_benchmark_ipc` is a jstp error, which will be fixed by @belochub.
As for now - do not open to many connections per a client, 10 seems to be ok.

lundibundi:
Adding OOM error screenshots just for reference.
- For ipc connectinon:
deepinscreenshot_20170719172807
- For net socket:
deepinscreenshot_select-area_20170719172307

benchmark/run.js
```diff
@@ -0,0 +1,137 @@
...
+    Requests sent:                    ${count}
+    Mean time of one request:         ${mean * 1e-6} (ms)
+    Stdev of time of one request:     ${stdev * 1e-6} (ms)
+    Estimated RPS:                    ${erps}
+  `);
+  process.exit(0);
+}
+
+function terminate() {
+  console.log(`
```
lundibundi: As I see those line-breaks were intentional, but maybe it's better to just use '\n'?

benchmark/worker
```diff
@@ -0,0 +1,95 @@
+'use strict';
```
lundibundi: No `.js` in filename.

benchmark/run.js
belochub: @lundibundi, how is that better?
As I see this, both of this options are completely equal, meaning that it is just a style preference, and we don't have a linter rule for this case, thus it is okay to leave it this way.

lundibundi: Yep, I didn't insist on making it that way.

nechaido:
> So what about allowing to specify either filename or port to be able to test both ipc and net?

I don't like the idea, because it will only show, potential performance of JSTP over TCP. If we decide to allow user specify transport - we should allow to choose any transport.

I'd like to leave this PR this way, and add such a feature in future.

benchmark/run.js
nechaido: @lundibundi 
IMHO
```
  console.log(`
Benchmark is being terminated due to an error or signal termination
`);
```
looks better then:
```
  console.log(
    '\nBenchmark is being terminated due to an error or signal termination\n'
);
```

lundibundi: This one is even better in my opinion:
```
const message =
  '\nBenchmark is being terminated due to an error or signal termination\n';
console.log(message);
```

nechaido:
@belochub ping.

benchmark/run.js
```diff
@@ -0,0 +1,137 @@
...
+    .forEach((worker) => {
+      worker.kill('SIGKILL');
+    });
+
+  if (!serverExited) {
+    server.kill('SIGINT');
+    setTimeout(() => {
+      if (!serverExited) {
+        server.kill('SIGKILL');
+        console.log('Master processwas not able to close server gracefully');
```
belochub: Missing a space here.

nechaido:
@belochub @aqrln ping.

benchmark/run.js
```diff
@@ -0,0 +1,140 @@
...
+);
+
+const {
+  workers: workersAmount,
+  connections: connectionsPerWorker,
+  requests: requestsPerConnection,
+  size: argumentSize,
+} = args;
+
+
```
aqrln: Nit: extra line break.

benchmark/run.js
```diff
@@ -0,0 +1,140 @@
...
+  [workersAmount * connectionsPerWorker],
+  { stdin: 'pipe' }
+);
+
+let serverExited = false;
+
+const workers = new Array(workersAmount);
+const workersExited = new Array(workersAmount);
+
+const results = new Array(workersAmount);
```
aqrln: I don't care much, but it feels like this line and the previous one should have been swapped.

benchmark/server.js
```diff
@@ -0,0 +1,40 @@
...
+    method(connection, argument, callback) {
+      callback(null);
+    },
+  },
+});
+
+const server = jstp.net.createServer([app]);
+server.maxConnections = maxConnections;
+
+const socket = '/tmp/jstp_benchmark_ipc';
```
aqrln: This won't work on Windows at all (where named pipes are created inside a special namespace in the VFS), and maybe on some Unix flavors too (it isn't correct to rely on `/tmp/` being available, that's what `os.tmpdir()` is for). I am okay with landing this as is now, because the benchmark is not exposed to end users, but it should be fixed later. Aside from that, a hardcoded filename is prone to race conditions.

benchmark/worker.js
```diff
@@ -0,0 +1,98 @@
...
+    argument = '0'.repeat(payload[2]);
+  } else if (type === 'start') {
+    start(payload[0]);
+  }
+});
+
+function connect(socket) {
+  let connected = 0;
+  const createConnection = (index) => {
+    jstp.net.connectAndInspect('app', null, ['iface'], socket,
```
aqrln: Nit: I'd put the arguments on a new line.

benchmark/worker.js
```diff
@@ -0,0 +1,98 @@
...
+          process.send(['connected']);
+        }
+      }
+    );
+  };
+
+  for (let i = 0; i < connections.length; i++) {
+    createConnection(i);
+  }
+
```
aqrln: Nit: unnecessary empty line.

benchmark/worker.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function start(requests) {
+  const responseTimesHR = new Array(connections.length);
+  for (let i = 0; i < connections.length; i++) {
+    responseTimesHR[i] = new Array(requests);
+  }
+  let responses = 0;
+  let startTimeHR = new Array(2);
```
aqrln: Can you please remove `= new Array(2)` (or assign the variable a null value)?

benchmark/server.js
belochub: @aqrln, what kind of race conditions?

nechaido:
@belochub ping.

benchmark/run.js
```diff
@@ -0,0 +1,139 @@
...
+      worker.kill('SIGKILL');
+    });
+
+  if (!serverExited) {
+    server.kill('SIGINT');
+    setTimeout(() => {
+      if (!serverExited) {
+        server.kill('SIGKILL');
+        console.log('Master process was not able to close server gracefully');
+        fs.unlinkSync('/tmp/jstp_benchmark_ipc');
```
belochub: You forgot to change the path here.

benchmark/run.js
```diff
@@ -0,0 +1,141 @@
...
+let serverExited = false;
+
+const workers = new Array(workersAmount);
+const workersExited = new Array(workersAmount);
+
+const results = new Array(workersAmount);
+let workersConnected = 0;
+let workersFinished = 0;
+
+let becnhStartedHR;
```
belochub: Typo in word "bench" here.

benchmark/run.js
```diff
@@ -0,0 +1,141 @@
...
+    .forEach((worker) => {
+      worker.kill('SIGKILL');
+    });
+
+  if (!serverExited) {
+    server.kill('SIGINT');
+    setTimeout(() => {
+      if (!serverExited) {
+        server.kill('SIGKILL');
+        console.log('Master process was not able to close server gracefully');
```
belochub: It'll be better to replace `console.log` with `console.warn` here.

benchmark/server.js
aqrln: multiple benchmarks run simultaneously

belochub:
Landed in https://github.com/metarhia/jstp/commit/5bf394454cdd1212629b12fa815395628a875afe.

PR #254, opened 2017-07-12
lib,parser: optimize parser and socket transport

belochub:
* Optimize parser by using Buffers without conversion to string when possible and avoiding calling ToString when parsing strings.
* Change socket to work with Buffers instead of strings to avoid unneeded multiple conversion.
* Refactor (reimplement) parseNetworkPackets function to make it more verbose, improving readability, and change it to work on Buffers instead of strings.
* Avoid excessive copying in cases when messages are divided and being received by parts (experimental).

This PR somewhat experimental in some parts (mostly `lib/socket.js` changes) and is open to comments with better ideas on how to improve it.

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -62,25 +62,34 @@ class Transport extends EventEmitter {
   _onSocketData(chunk) {
     const messages = [];
     let parsedLength;
+    let newPartLength;
 
     if (this._buffer) {
-      this._buffer = Buffer.concat([this._buffer, chunk],
-        this._buffer.length + chunk.length);
-    } else {
-      this._buffer = chunk;
+      newPartLength = chunk.indexOf(0);
+      if (newPartLength === -1) {
+        this._buffer = Buffer.concat(
+          [this._buffer, chunk],
+          this._buffer.length + chunk.length
+        );
+        this._checkBufferSize();
+        return;
+      }
+      if (!this._parseRemains(chunk.slice(0, newPartLength), newPartLength)) {
+        return;
+      }
+      chunk = chunk.slice(newPartLength + 1);
+      this._buffer = null;
     }
 
     try {
-      parsedLength = serde.parseNetworkMessages(this._buffer, messages);
+      parsedLength = serde.parseNetworkMessages(chunk, messages);
     } catch (error) {
       this.socket.destroy(error);
       return;
     }
 
-    if (this._buffer.length === parsedLength) {
-      this._buffer = null;
-    } else {
-      this._buffer = this._buffer.slice(parsedLength);
+    if (chunk.length !== parsedLength) {
+      this._buffer = chunk.slice(parsedLength);
     }
 
     const messagesCount = messages.length;
@@ -88,7 +97,29 @@ class Transport extends EventEmitter {
       this.emit('message', messages[i]);
     }
 
-    if (this._buffer && this._buffer.length > MAX_MESSAGE_SIZE) {
+    if (this._buffer) {
+      this._checkBufferSize();
+    }
+  }
+
+  _parseRemains(newPart, newPartLength) {
+    let result;
+    const buffer = Buffer.concat(
+      [this._buffer, newPart],
+      this._buffer.length + newPartLength
+    );
+    try {
+      result = serde.parse(buffer);
+    } catch (error) {
+      this.socket.destroy(error);
+      return false;
+    }
+    this.emit('message', result);
+    return true;
+  }
+
+  _checkBufferSize() {
+    if (this._buffer.length > MAX_MESSAGE_SIZE) {
       this.emit('error', new Error('Maximal message size exceeded'));
     }
   }
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -19,10 +19,9 @@ class Transport extends EventEmitter {
     super();
 
     this.socket = socket;
-    this._buffer = '';
+    this._buffer = null;
     this._uncorkSocket = this.socket.uncork.bind(this.socket);
 
-    this.socket.setEncoding('utf8');
     this.socket.on('data', this._onSocketData.bind(this));
 
     common.forwardMultipleEvents(this.socket, this, ['error', 'close']);
@@ -62,13 +61,26 @@ class Transport extends EventEmitter {
   //
   _onSocketData(chunk) {
     const messages = [];
-    this._buffer += chunk;
+    let parsedLength;
+
+    if (this._buffer) {
+      this._buffer = Buffer.concat([this._buffer, chunk],
+        this._buffer.length + chunk.length);
+    } else {
+      this._buffer = chunk;
+    }
 
     try {
-      this._buffer = serde.parseNetworkMessages(this._buffer, messages);
+      parsedLength = serde.parseNetworkMessages(this._buffer, messages);
     } catch (error) {
       this.socket.destroy(error);
       return;
+    }
+
+    if (this._buffer.length === parsedLength) {
+      this._buffer = null;
+    } else {
+      this._buffer = this._buffer.slice(parsedLength);
     }
 
     const messagesCount = messages.length;
@@ -76,7 +88,7 @@ class Transport extends EventEmitter {
       this.emit('message', messages[i]);
     }
 
-    if (this._buffer.length > MAX_MESSAGE_SIZE) {
+    if (this._buffer && this._buffer.length > MAX_MESSAGE_SIZE) {
       this.emit('error', new Error('Maximal message size exceeded'));
     }
   }
--- a/src/message_parser.cc
+++ b/src/message_parser.cc
@@ -15,6 +15,7 @@ using std::size_t;
 using std::strlen;
 
 using v8::Array;
+using v8::Integer;
 using v8::Isolate;
 using v8::Local;
 using v8::String;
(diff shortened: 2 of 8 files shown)


lib/socket.js
```diff
@@ -62,20 +61,64 @@ class Transport extends EventEmitter {
...
   _onSocketData(chunk) {
     const packets = [];
-    this._buffer += chunk;
+    let parsedLength;
+    let newPartLength;
+
+    if (this._buffer) {
+      newPartLength = chunk.indexOf(0);
+      if (newPartLength === -1) {
+        this._buffer = Buffer.concat(
```
lundibundi: Maybe it's worth trying to allocate a big buffer and manually append data to it if there is enough space or if there is not enough space then copy both of them into a new bigger buffer till it's size grows to `MAX_PACKET_SIZE` (same approach may be used in `_parseRemains` to avoid creating a new buffer 2 more times)? Or maybe there were problems with this approach?

lib/socket.js
```diff
@@ -62,20 +61,64 @@ class Transport extends EventEmitter {
...
     for (let i = 0; i < packetsCount; i++) {
       this.emit('packet', packets[i]);
     }
 
+    if (this._buffer) {
+      this._checkBufferSize();
+    }
+  }
+
+  _parseRemains(newPart, newPartLength) {
```
lundibundi: Maybe name it `_parseWith` or `_parseWithChunk`?

lib/socket.js
belochub: What do you mean by "append"? Isn't it just copying from one buffer into another?

lib/socket.js
belochub: What do you think about `_parseMessageRemains`?

lib/socket.js
lundibundi: Yes but AFAIK `concat` always creates a new buffer to copy the ones passed. I propose to avoid copying both buffers if the first one contains enough space to fit the second buffer and just copy the second buffer to the first one, hence 'append'.

lib/socket.js
lundibundi: I do not understand the reason for "Remains", as I see this function actually parses the main part and what remains is the latter part of `chunk` after slice.

lib/socket.js
belochub: The size of the buffer will always be equal to the size of the data inside it because it is coming from 'data' event on the socket, so there is never enough space to copy the second buffer inside it. And you can't just resize the buffer because it is just a view over the ArrayBuffer underneath, you can only create a new one.
I still can't understand how you want this to be implemented.

lib/socket.js
belochub: "Remains" means message part that remains in the buffer from the previous parsing.

lundibundi: Ain't that function parse data that was accumulated (and including the next message) till we've received message with '\0' in it?

lib/socket.js
lundibundi: I meant this for a case where we receive consecutive messages without '\0' in them, so when first such message arrives we copy its contents to the buffer in this class that has enough space (with space to spare) or if it doesn't then we replace it with a bigger one and copy, hence on a second and so on half-messages we won't have to copy what we've received before and only append the received data to the buffer we have in class.

belochub: And how to determine the amount of spare space on this buffer creation?

lib/socket.js
belochub: No, it just parses one message combining new part of it with the remains in the buffer.

lib/socket.js
lundibundi: At first start with something like 100 and then use basic [growth factors](https://en.wikipedia.org/wiki/Dynamic_array) as 1.5 or 2?

lib/socket.js
lundibundi: But those 'remains' are actually what was accumulated while we've been waiting for the message with '\0', aren't they?

lib/socket.js
belochub: Why exactly have you chosen 100 and what does growth factor have to do with this at all?

lib/socket.js
belochub: It's just a message part that remained from the previous parsing.

lib/socket.js
lundibundi: No particular reason, I thought that should be enough for most of the messages at first. You asked for how to determine spare space, hence I proposed to just use growth factor to grow that class level buffer.

lib/socket.js
lundibundi: Ok, that's up to you anyway.

belochub: I don't understand this naming, `_parseRemainsWith` what? There should be a word after "with".

lib/socket.js
belochub: You never mentioned this buffer growing, when do you want it to grow?

lundibundi: I mentioned it in this [comment, third line](https://github.com/metarhia/jstp/pull/254#discussion_r127193334). It'll grow (be replaced with a bigger one) when message that has arrived cannot fit to the current one.

lib/socket.js
lundibundi: As I see it, first argument to the function is what `with` refers to.

lib/socket.js
belochub: And why will this be better?

lib/socket.js
belochub: Yes, but when you call this function, you can't tell what was the name of the arguments.

lib/socket.js
lundibundi: It'll avoid copying original message upon each of the consecutive received messages and will only append the received one to the buffer which is obviously less work.

lib/socket.js
lundibundi: Oh, well, I can 😄. Anyway my thought was that you don't need names of the arguments, it's that the arguments that you pass are what `with` refers to.

lib/socket.js
belochub: @lundibundi, from my observations, it is less common to receive long messages that will be divided more than once. And it is quite difficult to pick a starting size value for the buffer which will lead to improvement instead of being an unneeded overhead, I think it should be determined in an empirical way.
Also, my question in https://github.com/metarhia/jstp/pull/254#discussion_r127284438 was about this https://github.com/metarhia/jstp/pull/254#discussion_r127280400 (my bad, wasn't specific enough), meaning I was asking about the buffer growth: as far as I understand, it will lead to the same copying of the initial buffer that we have right now, and I don't understand how this will be better for performance. Can you, please, explain this part of your idea exactly?

lundibundi: @belochub surely, empirical way will be better, I just suggested a starting point to bench different values. Also the way I propose should only kick in if we receive half-message and in case we receive full message we should not copy it to the intermediate buffer.
Regarding the idea: now when we receive first half-message we concat it with null(initial state of class buffer), upon each next half-message we receive we will copy current class buffer and received data to the new bigger buffer with the size exactly the sum of the previous' buffer sizes. Hence every time we receive a message we will copy every data we've received so far to the new buffer and then copy the received data. Please correct me if I see it wrong.
With the way I propose to do it we will copy the first half-message to the class buffer (which will always exist of some size), then upon receiving consequent half-messages we will append them to that class buffer hence only coping the received data. This way we will only copy our accumulated data if there is not enough space in the class buffer (and we will grow it more that just the size of received data), hence avoiding a lot of copying in case of very fragmented data.

belochub: @lundibundi

> now when we receive first half-message we concat it with null(initial state of class buffer)

Where does this happen exactly? I don't recall this behavior at all.

> upon each next half-message we receive

There literally can only be two half-messages :)

> then upon receiving consequent half-messages

Same here, can you please explain what you mean by "consequent half-messages"?

> This way we will only copy our accumulated data if there is not enough space in the class buffer (and we will grow it more that just the size of received data), hence avoiding a lot of copying in case of very fragmented data.

Yes and I was talking exactly about the growing part, it will lead to almost the same amount of copying we have now (see first paragraph in https://github.com/metarhia/jstp/pull/254#discussion_r128333507), and will lead to increased memory usage, I don't understand how this will be better for performance overall.

lundibundi: Oh yeah, we doesn't concat with null, we just assign it to _buffer.

If we assume that 
> There literally can only be two half-messages :)

Then yes, there will be the same amount of copying.
If we don't need to handle the case where there are more that 2 half-messages then my suggestion will only make the code more complex, so we should not implement it in this case.

belochub: @lundibundi, there is nothing to assume, it is a fact that there can't be more than two halves of a whole message (half-messages).

nechaido: @belochub I think that @lundibundi meant partial-message not half-message. He used the wrong term for it.

belochub: @nechaido, okay then, to benchmark this approach we will need an option to set the message size in benchmark, can you, please, add this option to the benchmark introduced in https://github.com/metarhia/jstp/pull/253?

nechaido:
@belochub could you rebase this on master and run benchmark with and without proposed changes and publish results here.

belochub:
Closing this, since serde implementation was moved from this package into the separate `mdsf` package.
Some of the optimizations from here should probably be moved into the `mdsf` package.

PR #255, opened 2017-07-14
test: add handshake timeout test

nechaido:


test/node/connection-handshake.js
```diff
@@ -126,3 +126,17 @@ test.test('must not accept handshakes on a client', (test) => {
   const connection = new jstp.Connection(transport, null, {});
   transport.emitPacket(handshake);
 });
+
+test.test(
```
lundibundi: There is no check that the connection was dropped after `HANDSHAKE_TIMEOUT` and exactly because of the timeout and not some kind of error. I think it's better to use 'handshakeTimeout' server event and check the `id` of a dropped connection. We may also add time checking (that it happened after time >`HANDSHAKE_TIMEOUT`) but that may not be necessary in my opinion.

nechaido: @lundibundi I've added check that `HANDSHAKE_TIMEOUT` occurs, but there is no reason in checking id of a connection, because there is only one connection.

lundibundi: @nechaido That's because there are no parallel test now, but there may be in the future, but that's up to you, I don't have a strong opinion on it.

nechaido:
Landed in cfbb4e9e38e58bbeb6e2735c2505e225dbda4d3a.

PR #258, opened 2017-07-19
test: enable tests for js parser implementation

aqrln:
The TODO tests were only launched using the default parser
implementation (which should be the native one while developing locally,
and is always the native one on CI since it fails the build if it
doesn't succeed building the addon).  This commit makes the tests run
with both implementations, the default and the fallback JavaScript-only
one.

### DIFF
--- a/test/node/serde.js
+++ b/test/node/serde.js
@@ -3,6 +3,7 @@
 const test = require('tap').test;
 
 const jstp = require('../..');
+const jsParser = require('../../lib/serde-fallback');
 
 const testCases = require('../fixtures/serde-test-cases');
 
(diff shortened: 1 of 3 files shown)


aqrln:
@nechaido absolutely right, but that'll be done when the JS parser implementation passes those tests (it currently does not).

belochub:
@aqrln, I think that JS parser does not pass this tests because they were not exactly correctly added to common serde tests, while they should've been added to the todo tests in the first place.
In my opinion, you should still move the tests, failing on JS parser, to the todo category and enable common serde tests to run on both parsers every time like @nechaido proposed. That way whenever you will implement a feature from a todo tests list, you will move some of the tests from todo to common tests, and common tests will indicate that you haven't broken anything that worked before the patch.

aqrln:
Sorry, I was confused. The JS parser passes the non-todo tests successfully.

test/node/serde.js
```diff
@@ -3,6 +3,7 @@
 const test = require('tap').test;
 
 const jstp = require('../..');
+const jsParser = require('../../lib/record-serialization-fallback');
```
belochub: This file has another name on the `master` branch now.

aqrln: Thanks, updated.

aqrln:
Landed in 2054024d8af705d40c8b130c0ae14bc3b154cfd6.

PR #259, opened 2017-07-20
tools: disable history for `readline.Interface`

belochub:
Disable history for `readline.Interface` used in Unicode tables generator thus improving overall performance and memory usage.


belochub:
Landed in https://github.com/metarhia/jstp/commit/26e64dcc5e2c82f78de348cd7f1584291e3a89f6.

PR #260, opened 2017-07-21
test: fix race condition in connection-handshake

nechaido:


test/node/connection-handshake.js
```diff
@@ -136,11 +136,11 @@ test.test(
     });
 
     const port = server.address().port;
-    net.connect(port, (error) => {
-      test.assertNot(error, 'must connect to server');
-      server.getClients()[0].on('close', () => {
-        test.pass('connection must be closed');
-      });
+    const connection = net.connect(port, error =>
+      test.assertNot(error, 'must connect to server')
```
aqrln: Nit: this function should have braces.

aqrln:
Strictly speaking, this doesn't have anything to do with macOS per se, it's just the bug hasn't been reproduced on Linux, so perhaps something like "test: fix race condition in connection-handshake" might be a better commit message.

Aside from that, the fact that it was only failing on macOS consistently makes it apparent that we need better platform coverage in CI, i.e., we should add macOS to the Travis configuration matrix and start using AppVeyor to test code on Windows.

aqrln:
I'd want to fast-track this fix. Without the patch, tests are currently failing on master for me.

belochub:
Landed in https://github.com/metarhia/jstp/commit/90c064c16fcb11622ac96b6f4be5277805f23519.

PR #264, opened 2017-07-22
src: fix building on the old gcc versions

belochub:


aqrln:
This is okay (and even preferable) to land without `.travis.yml` changes right away not waiting for 24 hours.

belochub:
Landed in https://github.com/metarhia/jstp/commit/59abe280c4c83b43d3f67743a89157d95d904ccb.

PR #265, opened 2017-07-22
build: fix building with Node.js 6 on macOS

aqrln:
Node.js 6.11.1 bundles npm 3.10.10, which bundles node-gyp 3.4.0.  The
problem with this version is that it doesn't emit `-stdlib=libc++` as
one of clang flags by default, which causes it to use old GCC 4.2
headers instead of clang headers.  Because of that, compilation fails
with:

    ../src/parser.cc:10:10: fatal error: 'cstdint' file not found

This patch adds the flag to `OTHER_CPLUSPLUSFLAGS`.  It causes no harm
when using newer node-gyp because duplicate flags are ignored, but
enables building JSTP with vanilla Node.js 6.x on macOS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,16 @@ node_js:
   - "7.10"
   - "8.0"
   - "8.1"
+matrix:
+  include:
+    - os: osx
+      osx_image: xcode8.3
+      node_js: 6
+      env:
+    - os: osx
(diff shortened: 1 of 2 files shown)


aqrln:
Landed in 5ed6bf0755ecd6d8308dcd84688549b6addcc31c.

PR #266, opened 2017-07-23
lib: add w3c websocket support to forwardEvent

nechaido:


lib/common.js
```diff
@@ -7,8 +7,14 @@
 //   newEventName - name of the forwarded event (optional)
 //
 const forwardEvent = (from, to, eventName, newEventName = eventName) => {
-  from.on(eventName, (...eventArgs) => {
-    to.emit(newEventName, ...eventArgs);
+  if (from.on) {
+    from.on(eventName, (...eventArgs) => {
+      to.emit(newEventName, ...eventArgs);
```
tshemsedinov: Why do you checking `from.on` but not checking `to.emit`?
Do you trust message destination more than message source?

nechaido: Yes, because it is an internal function and destination is always provided by us, so we can control if it is an EventEmmiter while source is not e.g. W3CWebsocket.
Anyway this PR will be completely rewritten.

PR #267, opened 2017-07-24
test: use tap's built-in todo test facilities

aqrln:
tap supports marking tests as pending.  It still runs them but if they
do not pass they are displayed as purple instead of red and do not fail
the build.  Remove `test-todo` npm script, move `test/todo/json5.js` to
`test/node/` and delete `test/todo/serde.js` completely (instead, merge
its test cases together with mainline test cases and introduce a `todo`
property there).


aqrln:
Landed in 6a09d446b15d228ebfe01d06742246526a0ff8cf.

PR #269, opened 2017-07-25
lib: refactor remoteProxy using classes

nechaido:


lib/remote-proxy.js
```diff
@@ -1,69 +1,64 @@
 'use strict';
 
-const events = require('events');
-const util = require('util');
+const EventEmitter = require('events').EventEmitter;
```
belochub: You should use destructuring assignment here to keep the style consistent.

lib/remote-proxy.js
```diff
@@ -1,69 +1,64 @@
...
-// Emit local event.
-//   eventName - name of an event
-//   eventArgs - array of event arguments
-//
-RemoteProxy.prototype._emitLocal = function(eventName, eventArgs = []) {
-  events.EventEmitter.prototype.emit.call(this, eventName, ...eventArgs);
-};
 
-// Create a method in a remote proxy that will call the corresponding remote
-// method. This is implemented as a static method rather than an instance
```
belochub: > This is implemented as a static method rather than an instance	method so that it will not be rewritten by a remote API method with the same	name.

Why did you change it?
I think that reasoning here is clear and it should still be a static method after refactoring.

nechaido: @belochub it is only called in the constructor, so even it is going to be rewritten nothing bad will happen, also it is very unlikely for a public api to have method to start with an underscore.
@aqrln what do you think about this?

nechaido: @aqrln ping.

aqrln: It is okay to make it an instance method. Nothing bad will actually happen even if it gets overwritten. The possibility of having `emit` and `on` methods in an interface is much more concerning.

nechaido:
Landed in de38e27.

Issue #91, opened 2017-02-28
Incorrect term usage (s/packet(s)/message(s))?


aqrln:
It is not technically correct to say "JSTP packets" since packets are the PDUs of layer 3 protocols. Layer 5 protocols don't have a standardized concept of PDU (and what they transfer is simply referred to as data) but in practice their data units are commonly called messages.

aqrln:
ping @metarhia/jstp-core

belochub:
Yes, I agree with that, we should change this misuse of terns.

lundibundi:
Completely agree on this one, with this https://github.com/metarhia/jstp-java/commit/8671892d4ece4e983a952ae61d0bb2e5a31fe563 java version uses 'message' throughout its codebase.

PR #270, opened 2017-07-25
lib,src: rename term packet usages to message

lundibundi:
At last we've got to this. This was done mostly automatically so there may be some problems, feel free to comment.

Fixes: https://github.com/metarhia/jstp/issues/91

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -307,8 +307,8 @@ class Connection extends EventEmitter {
     this.emit('error', error, this);
   }
 
-  // Process parsed incoming messages
-  //   messages - array of messages
+  // Process parsed incoming message
+  //   message - parsed incoming message
   //
   _processMessage(message) {
     const keys = Object.keys(message);
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -10,10 +10,10 @@ const RemoteProxy = require('./remote-proxy');
 
 let nextConnectionId = 0;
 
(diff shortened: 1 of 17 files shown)


belochub:
@lundibundi, what about the `doc` folder? You haven't renamed the term there.

lundibundi:
@belochub That folder is not relevant and will be removed, so there is no need to waste time dong that.

belochub:
@lundibundi, we can't just remove the whole folder, there are files that are still relevant, for example, in the `meetings` folder.
And if we are going to remove them, why can't we just do this now?

lundibundi:
@belochub yeah, I'll change the meetings' files. Thanks.

belochub:
@lundibundi, I don't think that it is a good idea to change the old meeting notes.
In my opinion, it is better to just add a note about renaming the term.

lib/connection.js
```diff
@@ -303,67 +307,67 @@ class Connection extends EventEmitter {
     this.emit('error', error, this);
   }
 
-  // Process parsed incoming packets
-  //   packets - array of packets
+  // Process parsed incoming messages
+  //   messages - array of messages
```
belochub: The argument is singular, not plural, and it is not an array.

belochub:
Landed in https://github.com/metarhia/jstp/commit/726f8410badc36f7647f2b882a6c90f01240e7c0.

PR #277, opened 2017-07-27
doc: start working on new documentation

aqrln:
Remove the outdated docs and start working on the new documentation.

PR #280, opened 2017-07-28
deps: add missing uuid dependency

aqrln:
Add a missing dependency to package.json.

The package was located in the tree as a deduplicated transitive
dependency, so it went unnoticed until the tree changed its shape in
https://github.com/metarhia/jstp/pull/277


aqrln:
I am going to fast-track this because current version in `master` doesn't even work without devDependencies.

aqrln:
Landed in 5b398d0608543a94341f538aa59da85acbb15121.

PR #284, opened 2017-08-01
lib: fix server's `connect` event

belochub:
Emit `connect` event on the server only after a successful handshake and emit `handshakeRequest` on the server instead of the connection object.


belochub:
Landed in https://github.com/metarhia/jstp/commit/37aa8e9248eb77a4ef8113aa5f00124f5e08b2c0.

PR #285, opened 2017-08-01
lib: disable update of messageId on callback send

belochub:


belochub:
Landed in https://github.com/metarhia/jstp/commit/cd8e043b8fd72ccf12408b439f2ef8cb25144b14.

PR #286, opened 2017-08-01
lib: allow to set event handlers in application

lundibundi:
As the name states you can now set event handlers in the application itself, this will gradually improve server side code as you won't need to manually set `event` handlers on each connection.

### DIFF
--- a/lib/applications.js
+++ b/lib/applications.js
@@ -13,16 +13,21 @@ const errors = require('./errors');
 //   api - object that contains interfaces as its fields each of which
 //         contains functions by method names. Each method has the following
 //         signature (connection, <0 or more method arguments>, callback)
-//   version - application version of 'name' application
+//   eventHandlers - object that contains interfaces as its fields each
+//                   of which contains functions by event names. Each method
+//                   has the following signature
+//                   (connection, <0 or more event arguments>)
+//   version - application version of 'name' application (optional)
 //
 class Application {
-  constructor(name, api, version) {
+  constructor(name, api, eventHandlers, version) {
     [this.name, this.version] = common.rsplit(name, '@');
     if (!this.version) this.version = version;
     if (this.version && !semver.valid(this.version)) {
       throw new TypeError('Invalid semver version');
     }
     this.api = api;
+    this.eventHandlers = eventHandlers;
   }
 
   // Call application method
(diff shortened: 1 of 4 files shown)


lib/applications.js
```diff
@@ -60,6 +65,19 @@ class Application {
...
+
+  // Handle incoming event
+  //   connection - JSTP connection that received event
+  //   interfaceName - name of the interface
+  //   eventName - name of the event
+  //   args - event arguments
+  //
+  handleEvent(connection, interfaceName, eventName, args) {
+    if (this.eventHandlers) {
+      const handler = this.eventHandlers[interfaceName][eventName];
```
belochub: If there is no object available under `interfaceName` key, this line will throw, I think.

test/node/api-versions.js
```diff
@@ -30,8 +30,8 @@ const interfacesVLatest = {
   },
 };
 
-const appV1 = new jstp.Application(app.name, interfacesV1, '1.0.0');
-const appV2 = new jstp.Application(app.name, interfacesV2, '2.0.0');
+const appV1 = new jstp.Application(app.name, interfacesV1, null, '1.0.0');
```
belochub: You should provide an empty object instead of `null` here.

test/node/api-versions.js
```diff
@@ -30,8 +30,8 @@ const interfacesVLatest = {
...
+const appV2 = new jstp.Application(app.name, interfacesV2, null, '2.0.0');
```
belochub: Ditto.

test/node/api-versions.js
```diff
@@ -223,7 +223,7 @@ test.test('must throw an error on invalid version in application name',
 test.test('must throw an error on invalid version in application parameter',
   (test) => {
     test.throws(() => {
-      new jstp.Application('app', interfacesV1, '__invalid__');
+      new jstp.Application('app', interfacesV1, null, '__invalid__');
```
belochub: Ditto.

test/node/application-event-handlers.js
```diff
@@ -0,0 +1,43 @@
...
+  if (server) server.close();
+  done();
+});
+
+test.test('must call event handler in application on remote event', (test) => {
+  const expectedName = 'name';
+
+  const eventHandlers = {
+    someService: {
+      name: (connection, name) => {
```
belochub: It'll be better to use ECMAScript 2015 syntax for methods here.

belochub:
Landed in https://github.com/metarhia/jstp/commit/bf26ce315dc9a150193a19bc0c2228ea919e52ff.

PR #287, opened 2017-08-01
lib: allow to broadcast events from server

lundibundi:
This PR also contains https://github.com/metarhia/jstp/pull/284 as it won't work properly without it and https://github.com/metarhia/jstp/pull/288.

### DIFF
--- a/test/node/server-broadcast-event.js
+++ b/test/node/server-broadcast-event.js
@@ -1,0 +1,65 @@
+'use strict';
+
+const test = require('tap');
+
+const jstp = require('../..');
+
+const app = require('../fixtures/application');
+
+const application = new jstp.Application(app.name, app.interfaces);
+const serverConfig = {
+  applications: [application],
+  authPolicy: app.authCallback,
+};
+
+let server;
+let port;
+
+test.beforeEach((done) => {
+  server = jstp.net.createServer(serverConfig);
+  server.listen(0, () => {
+    port = server.address().port;
+    done();
+  });
+});
+
+test.afterEach((done) => {
+  server.close();
+  done();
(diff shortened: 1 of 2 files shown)


test/node/server-broadcast-event.js
```diff
@@ -0,0 +1,62 @@
...
+  test.plan(12);
+
+  let connectionCount = 3;
+  const callback = () => {
+    if (--connectionCount === 0) {
+      server.broadcast(iface, eventName, ...eventArgs);
+    }
+  };
+
+  connectAndCheckEvent(test, iface, eventName, eventArgs, callback);
```
belochub: You can use for loop here with `connectionCount` iterations, this will make this test more flexible.

lundibundi: That's fine with me too.

nechaido: I'd use for loop here, and change plan to `test.plan(4*connections)`.

test/node/server-broadcast-event.js
```diff
@@ -0,0 +1,64 @@
...
+  server.close();
+  done();
+});
+
+test.test('must broadcast event to all connected clients', (test) => {
+  const iface = 'someService';
+  const eventName = 'broadcast';
+  const eventArgs = ['hello'];
+
+  test.plan(12);
```
nechaido: See https://github.com/metarhia/jstp/pull/287#pullrequestreview-53558804.

belochub:
Landed in https://github.com/metarhia/jstp/commit/13f494169d643fa4aa16d3572e10a1758b4fac18.

PR #288, opened 2017-08-01
lib: refactor server's getClients method

lundibundi:
* remove caching from old getClients
* rename current method to getClientsArray
* add getClients method that returns Iterator


belochub:
Landed in https://github.com/metarhia/jstp/commit/4c1325d8992cfde734c39f976bfd542caa6394ab.

PR #289, opened 2017-08-01
lib: implement sessions

belochub:
Refs: https://github.com/metarhia/jstp/issues/35


lib/connection.js
```diff
@@ -145,32 +151,56 @@ class Connection extends EventEmitter {
...
-      this.sessionId = sessionId;
-      if (callback) {
-        callback(error, sessionId);
-      }
-    };
+    if (isNewSession) {
+      this._callbacks[messageId] = (error, sessionId) => {
+        if (!error) {
+          if (login && password) {
+            this.username = login;
```
lundibundi: Is there any need in this field (I think it's enough to have it inside Session object)?

lib/connection.js
```diff
@@ -328,15 +362,22 @@ class Connection extends EventEmitter {
...
+
   // Send a JSTP message over this connection
-  //   message - a message to send
+  //   preparedMessage - a prepared message to send
+  //   message - a message object (used for development events thus optional)
   //
-  _send(message) {
-    const data = serde.stringify(message);
-    this.transport.send(data);
+  _send(preparedMessage, message) {
```
lundibundi: Why can't we make 2 send methods `_send(message)` and `_sendPrepared(preparedMessage)` and always parse prepared message for debug mode events in the latter method with the former method only calling the latter one after preparing a message? This will avoid a lot of boilerplate code and only affect debug mode performance which is negligible.

lundibundi: Also, I propose to avoid manual calling of `session._bufferMessage` method and instead add a boolean flag into the proposed above `_sendPrepared` method which will also remove some of the boilerplate code?

lib/transport-common.js
```diff
@@ -47,22 +49,25 @@ const newConnectFn = (
       client.application = new Application('jstp', {});
     }
     if (!client.connectPolicy) {
-      client.connectPolicy = new SimpleConnectPolicy().connect;
+      client.connectPolicy = new SimpleConnectPolicy(client.session).connect;
```
lundibundi: What is the use of the `client.session` parameter in `SimpleConnectPolicy`, ain't it better to just omit the parameters?

belochub: I forgot to change this line, thanks.

lib/connection.js
belochub: I don't like the idea of creating the function just for stringifying messages, the output of which will always be parsed just again while sending the message, that doesn't make sense to me.

lib/connection.js
belochub: I think, it's better to duplicate it because it doesn't use much more memory but at the same time helps to avoid long access to it, like `connection.session.username`.

lib/connection.js
lundibundi: It'll only be parsed for debug messages which will not affect the production code. Also, it'll avoid passing both parsed and not parsed arguments to the `_send` function.

lib/connection.js
lundibundi: Even though it indeed has virtually no cost, but I think that's overkill to copy it for that, you can always cache it before or use sort of accessor function to get it if you want to avoid long access.

lib/connection.js
belochub: Yes, it will avoid passing both parsed and not parsed arguments to the `_send` method by parsing the stringified message every time, which, as I've already said, makes no sense to me.

lundibundi: Well it's debug mode so it's fine for me, @metarhia/jstp-core?

nechaido: I am ok with both versions.

lib/session.js
```diff
@@ -0,0 +1,66 @@
...
+// Extends Map class and thus can be used to store the current session state
+// independently of connection.
+class Session extends Map {
+  //   connection - Connection class object
+  //   sessionId - used on client to represent session with corresponding id
+  //               which exists on the server
+  constructor(connection, sessionId) {
+    super();
+
+    this._isServer = !sessionId;
```
lundibundi: Maybe just use `Math.abs` below? IMO it'll be cleaner.

lib/server.js
```diff
@@ -44,15 +48,15 @@ const initServer = function(
...
-    if (!authPolicy) authPolicy = new SimpleAuthPolicy();
-    this.startSession = authPolicy.startSession.bind(authPolicy);
+  if (typeof authPolicy === 'object') {
+    authPolicy = authPolicy.authenticate.bind(authPolicy);
   }
+  this.authPolicy = authPolicy;
 
   this.clients = new Map();
 
+  this.sessions = sessionStorageProvider || new SimpleSessionStorageProvider();
```
lundibundi: Maybe `sessionsProvider`?

lib/server.js
```diff
@@ -80,6 +84,62 @@ class Server {
...
+    if (strategy !== 'anonymous') {
+      if (!this.authPolicy) {
+        callback(errors.ERR_AUTH_FAILED);
+        return;
+      }
+      this.authPolicy(
+        connection, application, strategy, credentials,
+        (error, username) => {
+          if (error || !username) {
+            callback(errors.ERR_AUTH_FAILED);
```
lundibundi: Maybe allow user to pass their own error codes? (like check if error.code and if it doesn't exist only then pass default errors.ERR_AUTH_FAILED?)

belochub: I think this has nothing to do with this PR, I didn't change this behavior.
But if you want, you can open another PR with this functionality after this one will be landed.

lib/server.js
belochub: What about `sessionsStorage`?

lundibundi: LGTM.

nechaido:
Landed in 37918fec691890572cc21ddd0d92fa514bc5800a.

nechaido:
@belochub please rebase on master.

belochub:
@nechaido, done.

lib/applications.js
```diff
@@ -18,16 +18,24 @@ const errors = require('./errors');
 //                   has the following signature
 //                   (connection, <0 or more event arguments>)
 //   version - application version of 'name' application (optional)
+//   sessionStorageProvider - provider for session storage (optional).
+//                            If provided will be used to store sessions
```
aqrln: I feel like the subject is missing in this sentence. How about "If provided, it will..."?

lib/applications.js
```diff
@@ -18,16 +18,24 @@ const errors = require('./errors');
...
+//                            independently of other applications
 //
 class Application {
-  constructor(name, api, eventHandlers = {}, version) {
+  constructor(name, api, eventHandlers = {}, version, sessionStorageProvider) {
+    if (typeof version === 'object') {
```
aqrln: I'd say something like `if (sessionStorageProvider === undefined && typeof version === 'object')` would be more robust.

FWIW, I'm not entirely happy with the very idea of omitting some arguments in the middle of the signature, but we already have such APIs in the library which I introduced myself in the past anyway, so ¯\\\_(ツ)\_/¯

lib/connection.js
```diff
@@ -145,32 +143,56 @@ class Connection extends EventEmitter {
     }
 
     let message;
+    let isNewSession = true;
+    const handshakePayload = new Array(2);
```
aqrln: If the whole purpose of this array is to eventually unpack it using spread operator, I'd actually find it more readable to use just two named variables (e.g., `handshakeStrategy` and `handshakeCredentials`).

lib/connection.js
```diff
@@ -294,56 +315,47 @@ class Connection extends EventEmitter {
...
   //
-  _send(message) {
-    const data = serde.stringify(message);
-    this.transport.send(data);
+  _send(preparedMessage, message) {
+    this.transport.send(preparedMessage);
 
     if (process.env.NODE_ENV !== 'production') {
-      this.emit('outgoingMessage', message);
+      this.emit('outgoingMessage', message || serde.parse(preparedMessage));
```
aqrln: `serde.parse()` is used directly here, but `serde.stringify()` is abstracted away into `_prepareMessage()`, which is a bit inconsistent.

lib/server.js
```diff
@@ -80,6 +81,78 @@ class Server {
     });
   }
 
+  startSession(connection, application, strategy, credentials, callback) {
+    const createSession = (username) => {
+      const session = new Session(connection);
+      if (username) {
+        session.username = username;
```
nechaido: I'd move it to the constructor.

lib/server.js
```diff
@@ -80,6 +81,78 @@ class Server {
     });
   }
 
+  startSession(connection, application, strategy, credentials, callback) {
+    const createSession = (username) => {
```
nechaido: I'd extract it to a private function that takes `(connection, application, strategy, callback)` as arguments.

lib/connection.js
```diff
@@ -145,32 +143,56 @@ class Connection extends EventEmitter {
...
-      }
-    };
+    if (isNewSession) {
+      this._callbacks[messageId] = (error, sessionId) => {
+        if (!error) {
+          if (login && password) {
+            this.username = login;
+          }
+          this.session = new Session(this, sessionId);
+          this.session.username = this.username;
```
lundibundi: Just a nit, but maybe it'll be better to make Session's constructor get `username` from the connection parameter instead of assigning it here manually?

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+    delete session.storage;
+    return Object.assign(result, session);
+  }
+
+  _bufferMessage(id, message) {
+    this.buffer.set(Math.abs(id), message);
+    this.latestBufferedMessageId = id;
+  }
+
+  _onMessageRecieved(messageId) {
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+    this.recievedCount = Math.abs(messageId);
+  }
+
+  _onCallbackMessageRecieved(messageId) {
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+    messageId = Math.abs(messageId);
+    for (let i = this.guaranteedDeliveredCount + 1; i <= messageId; i++) {
+      this.buffer.delete(i);
+    }
+    this.guaranteedDeliveredCount = messageId;
+  }
+
+  _restoreSession(newConnection, recievedCount) {
```
aqrln: rec**ie**ved → rec**ei**ved

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+    if (this.connection) {
+      this.connection.close();
+    }
+    this.connection = newConnection;
+    for (let i = this.guaranteedDeliveredCount + 1; i <= recievedCount; i++) {
```
aqrln: rec**ie**ved → rec**ei**ved

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+      this.buffer.delete(i);
+    }
+    this.guaranteedDeliveredCount = recievedCount;
```
aqrln: rec**ie**ved → rec**ei**ved

lib/connection.js
```diff
@@ -463,34 +476,66 @@ class Connection extends EventEmitter {
...
-    this.sessionId = sessionId;
 
-    this.emit('client', sessionId, this);
+    this.emit('client', session, this);
     this.server.emit('connect', this);
 
-    const message = this._createMessage('handshake', null, 'ok', sessionId);
-    this._send(message);
+    const message = this._createMessage(
+      'handshake', null, 'ok', session.recievedCount
```
aqrln: rec**ie**ved → rec**ei**ved

lib/connection.js
```diff
@@ -545,6 +590,8 @@ class Connection extends EventEmitter {
 
     const callback = this._remoteCallbackWrapper.bind(this, messageId);
 
+    this.session._onMessageRecieved(messageId);
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/connection.js
```diff
@@ -570,6 +617,7 @@ class Connection extends EventEmitter {
     const callback = this._callbacks[messageId];
 
     this.emit('callback', message.error, message.ok);
+    this.session._onCallbackMessageRecieved(messageId);
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/connection.js
```diff
@@ -595,6 +643,7 @@ class Connection extends EventEmitter {
     const eventArgs = message[eventName];
 
     this.emit('event', interfaceName, eventName, eventArgs);
+    this.session._onMessageRecieved(message.event[0]);
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/connection.js
```diff
@@ -614,6 +663,7 @@ class Connection extends EventEmitter {
     const interfaceName = message.inspect[1];
 
     this.emit('inspect', interfaceName);
+    this.session._onMessageRecieved(messageId);
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/connection.js
```diff
@@ -627,14 +677,18 @@ class Connection extends EventEmitter {
   //   message - parsed message
   //
   _processPingMessage(message) {
-    this.pong(message.ping[0]);
+    const messageId = message.ping[0];
+    this._pong(messageId);
+    this.session._onMessageRecieved(messageId);
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/connection.js
```diff
@@ -627,14 +677,18 @@ class Connection extends EventEmitter {
...
   }
 
   // Process incoming pong message
   //   message - parsed message
   //
   _processPongMessage(message) {
     const messageId = message.pong[0];
+    this.session._onCallbackMessageRecieved(messageId);
```
aqrln: Rec**ie**ved → Rec**ei**ved

lib/server.js
```diff
@@ -80,6 +81,78 @@ class Server {
...
+          callback(errors.ERR_AUTH_FAILED);
+          return;
+        }
+        createSession(username);
+      }
+    );
+  }
+
+  restoreSession(connection, application, credentials, callback) {
+    const [sessionId, recievedCount] = credentials;
```
aqrln: rec**ie**ved → rec**ei**ved

lib/server.js
```diff
@@ -80,6 +81,78 @@ class Server {
...
+      session = sessionStorageProvider.get(sessionId);
+    }
+
+    if (!session) {
+      callback(errors.ERR_AUTH_FAILED);
+      return;
+    }
+
+    this._clientsBySessionId.set(sessionId, connection);
+    session._restoreSession(connection, recievedCount);
```
aqrln: rec**ie**ved → rec**ei**ved

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+    super(state);
+
+    this.id = sessionId || uuid4();
+    this.connection = connection;
+    this.username = null;
+
+    this.guaranteedDeliveredCount = 0;
+    this.buffer = new Map();
+
+    this.recievedCount = 0;
```
aqrln: rec**ie**ved → rec**ei**ved

aqrln:
(the last 15 comments are the same typo/misspelling, I just highlighted all of the occurrences explicitly)

lib/session.js
```diff
@@ -0,0 +1,102 @@
...
+    this.latestBufferedMessageId = 0;
+
+    Object.preventExtensions(this);
```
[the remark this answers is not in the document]
belochub: To prevent users from creating new fields in session objects and thus enforce users to use session object's storage only by using methods inherited from Map.

lib/connection.js
```diff
@@ -294,56 +316,54 @@ class Connection extends EventEmitter {
   // Close the connection
   //
   close() {
+    this._closed = true;
```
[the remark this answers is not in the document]
belochub: What does it have to do with this PR?
This was added in https://github.com/metarhia/jstp/pull/303 by @nechaido.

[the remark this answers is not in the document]
belochub: +1 for the follow-up PR.

lib/server.js
```diff
@@ -20,7 +24,7 @@ const HANDSHAKE_TIMEOUT = 3000;
 //              server 'connect' event (optional).
 //
 const initServer = function(
```
lundibundi: You should add sessionStorageProvider parameter to the description of the method.

lib/server.js
```diff
@@ -36,22 +40,19 @@ const initServer = function(
...
   this.applications = applications;
   this.heartbeatInterval = heartbeatInterval;
 
-  if (typeof authPolicy === 'function') {
-    this.startSession = authPolicy;
-  } else {
-    if (!authPolicy) authPolicy = new SimpleAuthPolicy();
-    this.startSession = authPolicy.startSession.bind(authPolicy);
+  if (typeof authPolicy === 'object') {
+    authPolicy = authPolicy.authenticate.bind(authPolicy);
```
lundibundi: Maybe rename `authPolicy` to `authenticate`? It's usage in code will be clearer in my opinion.

lib/session.js
```diff
@@ -0,0 +1,103 @@
...
+
+  _onCallbackMessageReceived(messageId) {
+    messageId = Math.abs(messageId);
+    for (let i = this.guaranteedDeliveredCount + 1; i <= messageId; i++) {
+      this.buffer.delete(i);
+    }
+    this.guaranteedDeliveredCount = messageId;
+  }
+
+  _restoreSession(newConnection, receivedCount) {
```
lundibundi: I propose to rename this method to simply `_restore`, to avoid duplication.

lib/simple-connect-policy.js
```diff
@@ -18,14 +21,25 @@ module.exports = class SimpleConnectPolicy {
...
   //              (error, connection)
   //
-  connect(app, connection, callback) {
-    connection.handshake(
-      app, this.login, this.password,
-      (error) => {
-        callback(error, connection);
-      });
+  connect(app, connection, session, callback) {
+    if (!session) {
```
lundibundi: This if can be removed completely by assigning this.login to session parameter if `!session` and using handshake method with session and this.password and 2, 3 paraments. Also currently callback in a first if branch passes invalid session object (should be this.login or null).

lib/simple-session-storage-provider.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  // Must save the Session object with the corresponding session id.
+  // This method is called once on session creation and every time
+  // when connection associated with session is being closed.
+  //   sessionId - id of the session to be added
+  //   session - Session object to be added
+  //
+  set(sessionId, session) {
+    const storedSession = {
```
lundibundi: You can avoid creating new object when there is one in the map already and only change its properties here, but this is not a very 'hot' code so it's fine with me either way.

test/node/session-resend-event-on-connection-drop.js
```diff
@@ -0,0 +1,70 @@
...
+      session(...args);
+      break;
+  }
+});
+
+server.listen(0, () => {
+  client.send(['connect', server.address().port]);
+});
+
+function getClient() {
```
lundibundi: `createClient` is a better name here in my opinion.

test/node/session.js
```diff
@@ -0,0 +1,65 @@
...
+  test.plan(6);
+  port = server.address().port;
+  jstp.net.connect(app.name, null, port, (error, conn, session) => {
+    test.assertNot(error, 'handshake must not return an error');
+    test.equal(conn.username, null, 'username must be null');
+    test.assert(session instanceof jstp.Session,
+      'session must be an instance of jstp.Session');
+    const client = { session };
+    conn.callMethod('calculator', 'doNothing', [], (error) => {
+      test.assertNot(error, 'call must not return an error');
```
lundibundi: As I understand this test is based on multiple calls of this test.assertNot which will lead to test failure but this callback will not be called again after first call as it's a callback on a different connection.

test/utils/session/client.js
```diff
@@ -0,0 +1,57 @@
...
+    client.session = session;
+    connection.callMethod(iface, method, [], (error) => {
+      if (error) {
+        process.send(['error', error]);
+      }
+    });
+  });
+};
+
+const getSession = () => {
```
lundibundi: This method sends serialized session to master process but is called getSession, should probably be renamed to sendSession.

nechaido:
@aqrln ping.

belochub:
Landed in 4d600f4587238f0a8165bbb812e4570c5a9f01c7, fb8915cd435d2c24d23579075ec4b7770a9022eb, f9b888c68a575b958751636e38b9300f1409aee4 and
 8a0bc8bce910706f060432f1d65b5e827cdbb31d. 🎉.

PR #292, opened 2017-08-03
bench: add distributed benchmark

nechaido:


benchmark/distributed/master.js
```diff
@@ -0,0 +1,163 @@
...
+  })
+  .options('cert', {
+    alias: 'C',
+    describe: 'path to file that contains SSL certificate',
+    type: 'string',
+    coerce:
+      arg => (arg ? fs.readFileSync(path.resolve(__dirname, arg)) : undefined),
+  })
+  .help()
+  .alias('help', 'h')
```
lundibundi: Why lowercase, all other options are uppercased?
Also, I'd still like for all short options to be lowercase.

benchmark/distributed/master.js
```diff
@@ -0,0 +1,163 @@
+'use strict';
+
+const jstp = require('../..');
+
+const yargs = require('yargs');
+const fs = require('fs');
+const path = require('path');
+const parseAddress = require('./parse-address');
+const statistics = require('../statistics');
```
lundibundi: I'd to suggest to split local files from npm packets via empty line as we have in `lib`. I've checked the other benchmark and it also has it all in one place, so I think if we agree on this one we should change it there too.

benchmark/distributed/master.js
```diff
@@ -0,0 +1,163 @@
...
+  })
+  .options('report', {
+    alias: 'R',
+    describe: 'Report interval in ms',
+    type: 'number',
+    default: 10000,
+  })
+  .options('key', {
+    alias: 'K',
+    describe: 'path to file that contains SSL certificate key',
```
lundibundi: First letter style in `describe` should be consistent, either all fields have the first letter uppercased or all don't.

benchmark/distributed/parse-address.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
```
lundibundi: Btw, maybe extract this part of the [cli/connect](https://github.com/metarhia/jstp/blob/master/lib/cli/line-processor.js#L73) function and use it both here and in cli, it also has a few good errors?

benchmark/distributed/master.js
nechaido: `-h` is common short for `help`. Other options short variant are uppercased because it is a common practice, IMHO.

aqrln: @nechaido I'm not sure about the common practice. Virtually all CLI applications I've seen fall back to uppercase options only when they have a lot of them and run out of lowercase mnemonics.

benchmark/distributed/parse-address.js
nechaido: > also has a few good errors?

I don't get what you mean.

benchmark/distributed/master.js
lundibundi: Well I'm not sure about common practice as all Linux tools I've used have options in both cases (mostly because there are so many of them), but most of the options I use every day are lowercase so I get an impression that those are more common instead. But in any case, you've created this 'tool' so it's up to you to decide on case.

benchmark/distributed/parse-address.js
lundibundi: It checks host, port for existence and minimal validity.

nechaido: Now I get it. That means it has validation, but not errors.

benchmark/distributed/master.js
```diff
@@ -0,0 +1,163 @@
...
+      const requests = workersReportBuffer.reduce(
+        (sum, current) => sum + current, 0
+      );
+      workersReportBuffer = [];
+      workersReports.push([requests, process.hrtime(startTimeHR)]);
+      const rps = requests * 1e9 / timePassed;
+      rpsReports.push(rps);
+      const mean = statistics.mean(rpsReports);
+      const stdev = statistics.stdev(rpsReports, mean);
+      console.log(`Current RPS: ${rps}, mean: ${mean}, stdev: ${stdev}`);
```
lundibundi: Let's add passed time here?

nechaido: What for?

lundibundi: This will allow making a bit more reasonable decision of importance of this log, as it's also a variable value.

benchmark/distributed/master.js
```diff
@@ -0,0 +1,163 @@
...
+master.listen(listenAddress, () => {
+  console.log(`Master server listening on ${address}`);
+  startTimeHR = process.hrtime();
+
+  let intervalTimerHR = process.hrtime();
+  const rpsReports = [];
+  setTimeout(() => {
+    setInterval(() => {
+      const timePassedHR = process.hrtime(intervalTimerHR);
+      const timePassed = timePassedHR[0] * 1e9 + timePassedHR[0];
```
lundibundi: Ain't it `const timePassed = timePassedHR[0] * 1e9 + timePassedHR[1]`?

benchmark/distributed/master.js
```diff
@@ -0,0 +1,163 @@
...
+        connection.emitRemoteEvent('master', 'connect', [serverAddress]);
+      }
+
+      connection.on('event', (iface, event, args) => {
+        if (iface !== 'master') {
+          return;
+        }
+        if (event === 'report') {
+          const requests = +args[0];
+          workersReportBuffer.push(requests);
```
lundibundi: Why not just add the value to global counter here, is there any use of keeping them in the array?

benchmark/distributed/master.js
nechaido: IMHO, there sis no reason at all to show it here. It will always say 10±1 second.

benchmark/distributed/master.js
nechaido: Yep, this way in the future, we can save the data to a file, and then analyze it with some tool. We can't analyze RPS's because that value is not precise.

benchmark/distributed/worker.js
```diff
@@ -0,0 +1,127 @@
...
+  })
+  .option('interval', {
+    alias: 'i',
+    describe: 'Interval of requests sent by connection in ms',
+    type: 'number',
+    default: 10,
+  })
+  .option('size', {
+    alias: 's',
+    describe: 'Size of argument to call serverside method with',
```
lundibundi: server-side?

benchmark/distributed/worker.js
```diff
@@ -0,0 +1,127 @@
...
+const createConnection = (index, address) => {
+  jstp[transport].connectAndInspect(
+    'server', null, ['server'], ...address, (error, connection, api) => {
+      if (error) {
+        console.error(`Could not connect to server:\n${error}`);
+        return;
+      }
+
+      connections[index] = connection;
+      timers[index] = setInterval(() => {
```
lundibundi: I think it's better to set one interval timer and make all calls in there, this will most likely make clients more performant.

lundibundi: Btw let's also count server response times and report their mean?

lundibundi:
Btw maybe it'll be better to wait for the workers' report and upon every full report (every worker sent data) print the data (rps and such)?

benchmark/distributed/master.js
lundibundi: I didn't quite catch that, what does it have to do with RPS as this is just requests count? And what use will analyzer tools have for the array of requests counts from each worker over reportTime as it is being reset upon each report?

benchmark/distributed/master.js
lundibundi: Oh yeah, I forgot that master is idle most of the time.

benchmark/distributed/master.js
nechaido: Sorry, I mixed up, you are right.

nechaido:
@lundibundi this way (https://github.com/metarhia/jstp/pull/292#issuecomment-3218728140) you may wait for it for a long-long time.

benchmark/distributed/master.js
nechaido: Thankfully to https://github.com/metarhia/jstp/pull/292#discussion_r132741231, I now remember the purpose of this line, it is for implementing  https://github.com/metarhia/jstp/pull/292#discussion_r132741231.

lundibundi:
@nechaido That's a possibility but it should help stabilize the rps results when we report and record them without one or more reports from workers. Also, there is not much value in having the benchmark report in exactly reportTime, as it doesn't really matter, results significance is a lot more important.

nechaido:
@lundibundi proposed approach will not increase significance at all.

nechaido:
@metarhia/jstp-core ping.

benchmark/distributed/master.js
nechaido: Now every short option is lowercased.

benchmark/distributed/master.js
```diff
@@ -0,0 +1,193 @@
...
+    registerServer: (connection, address, callback) => {
+      if (server) {
+        callback('Server is already registered');
+      }
+      server = connection;
+      serverAddress = address;
+      workers.forEach((worker) => {
+        worker.emitRemoteEvent('worker', 'connect', [serverAddress]);
+      });
+      connection.on('event', (iface, event, args) => {
```
lundibundi: Why don't you use event handling api here (as well as in other places)?

benchmark/distributed/master.js
```diff
@@ -0,0 +1,193 @@
...
+        if (iface !== 'master') {
+          return;
+        }
+        if (event === 'report') {
```
lundibundi: Also, I think it's better to change event name for server and worker reports to be able to differentiate them clearly.

benchmark/distributed/master.js
```diff
@@ -0,0 +1,193 @@
...
+    const timePassedHR = process.hrtime(intervalTimerHR);
+
+    if (!workersReportBuffer.length) {
+      return;
+    }
+
+    const timeFromStart = timeFromStartHR[0] * 1e9 + timeFromStartHR[1];
+    const timePassed = timePassedHR[0] * 1e9 + timePassedHR[1];
+
+    const samples = workersReportBuffer.map(
```
lundibundi: Why not just split them on event handling? (As I see it you wanted to localize argument handling hence placed it here but I think it's better to handle them right away as this function has nothing to do with the format workers present their data with, also with new event handling api you will have all arguments already separated upon handling).

benchmark/distributed/server.js
```diff
@@ -0,0 +1,148 @@
...
+
+// Prevent Node.js from throwing DEPTH_ZERO_SELF_SIGNED_CERT.
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+
+const server = jstp[transport].createServer(config);
+server.maxConnections = 100000;
+
+server.listen(listenAddress, () => {
+  console.log(`Server listening on ${address}`);
+  startTimeHR = process.hrtime();
```
lundibundi: Why do you start measuring time here instead of after connection establishment and before `api.master.registerServer`? Or maybe it'll be better to start measuring after `registerServer` but on master-side emit `connect` event after (or setImmediate) calling callback of this method?

benchmark/distributed/worker.js
```diff
@@ -0,0 +1,160 @@
...
+
+      connections[index] = connection;
+    }
+  );
+};
+
+const report = () => {
+  const req = requests;
+  requests = 0;
+  const responseTimes = responseTimesHR.map(
```
lundibundi: Any use of calculating time here instead of immediately after receiving the callback?

benchmark/distributed/server.js
```diff
@@ -0,0 +1,148 @@
...
+      api.master.registerServer(connectAddress, (error) => {
+        if (error) {
+          console.error(error);
+          connection.close();
+          server.close();
+          return;
+        }
+
+        timer = setInterval(() => {
+          const timePassedHR = process.hrtime(startTimeHR);
```
lundibundi: Is my understanding correct that you wanted this to be a total time from the start of the server or this is a mistake?

benchmark/distributed/master.js
nechaido: Because this PR is older than those features. But thanks for pointing at it, I will rewrite it.

benchmark/distributed/server.js
nechaido: Yes, it is.

benchmark/distributed/server.js
nechaido: Because this timer is meant to represent time from passed from master server start.

benchmark/distributed/worker.js
nechaido: Yes. This way we can use library that will provide precise integer operations, when we find one. If you calculate time immediately after receiving callback precision will be lost.

nechaido:
Landed in 630920a.

PR #281, opened 2017-07-29
lib: add handling of EAGAIN error on connection

belochub:
This fixes the connection error in benchmark reported by @lundibundi in https://github.com/metarhia/jstp/pull/253#pullrequestreview-50054575.

PR #293, opened 2017-08-03
lib: fix multiple calls of callback in connect

belochub:
This fixes bug introduced in https://github.com/metarhia/jstp/pull/281.

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -86,15 +86,16 @@ const socketFactory = (connect) => {
   const resultConnectFn = (...options) => {
     const callback = common.extractCallback(options);
     const socket = connect(...options);
-    socket.once('error', (error) => {
+    const onceErrorListener = (error) => {
       if (error.code !== 'EAGAIN') {
         callback(error);
         return;
       }
       process.nextTick(resultConnectFn, ...options, callback);
-    });
+    };
+    socket.once('error', onceErrorListener);
     socket.once('connect', () => {
-      socket.removeListener('error', callback);
+      socket.removeListener('error', onceErrorListener);
       callback(null, socket);
     });
   };
--- a/test/node/connection-callback-multiple-calls.js
+++ b/test/node/connection-callback-multiple-calls.js
@@ -1,0 +1,22 @@
+'use strict';
+
+const tap = require('tap');
+const jstp = require('../..');
+
+const app = new jstp.Application('app', {});
+const server = jstp.net.createServer({ applications: [app] });
+
+tap.plan(1);
+
+server.listen(0, () => {
+  const port = server.address().port;
+  jstp.net.connect(app.name, null, port, (error, connection) => {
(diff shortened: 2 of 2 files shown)


test/node/connection-callback-multiple-calls.js
```diff
@@ -0,0 +1,22 @@
...
+
+server.listen(0, () => {
+  const port = server.address().port;
+  jstp.net.connect(app.name, null, port, (error, connection) => {
+    tap.pass('must call callback once');
+    connection.getTransport().destroy();
+    connection.on('error', () => {
+      // dismiss
+    });
+    connection.emitRemoteEvent('someService', 'someEvent', []);
```
nechaido: Why do you need  this line?

belochub: To get an error by writing after socket is destroyed, so that `connect` callback is called twice before the fix introduced in this PR.

nechaido:
@belochub please fix linter errors.

test/node/connection-callback-multiple-calls.js
```diff
@@ -0,0 +1,23 @@
...
+const app = new jstp.Application('app', {});
+const server = jstp.net.createServer({ applications: [app] });
+
+tap.plan(1);
+
+server.listen(0, () => {
+  const port = server.address().port;
+  jstp.net.connect(app.name, null, port, (error, connection) => {
+    tap.pass('must call callback once');
+    if (error) return;
```
nechaido: I'd prefer this to use assertion here. But if you like it his way - you can go ahead and merge.

test/node/connection-callback-multiple-calls.js
```diff
@@ -0,0 +1,37 @@
...
+    connection.on('error', () => {
+      // dismiss
+    });
+    connection.emitRemoteEvent('someService', 'someEvent', []);
+  });
+});
+
+test.test('must call connect callback once on error on connect', (test) => {
+  test.plan(1);
+  jstp.net.connect(app.name, null, 0, (error) => {
```
nechaido: It'sbetter to use:
```
const invalidAddress = {
  host: '__invalid_host__',
  port: 3000,
};
jstp.tls.connect(app.name, null, invalidAddress, (error) => {
```

lib/socket.js
```diff
@@ -86,15 +86,16 @@ const socketFactory = (connect) => {
   const resultConnectFn = (...options) => {
     const callback = common.extractCallback(options);
     const socket = connect(...options);
-    socket.once('error', (error) => {
+    const onceErrorListener = (error) => {
```
lundibundi: I'd prefer something like `errorHandler`, but that's up to you.

belochub:
Landed in https://github.com/metarhia/jstp/commit/c563b4cc221c7d418883ca7d0882498e9212fdff.

PR #268, opened 2017-07-24
test: add tests for websocket transport

nechaido:
Based on #266.

PR #295, opened 2017-08-08
test: rewrite ws, wss and W3C ws tests

nechaido:
Refs: https://github.com/metarhia/jstp/pull/268#issuecomment-318113152.

### DIFF
--- a/test/node/transport-w3c-ws.js
+++ b/test/node/transport-w3c-ws.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const test = require('tap');
+global.WebSocket = require('websocket').w3cwebsocket;
+const w3c = require('../../lib/ws-browser');
 
 const jstp = require('../..');
 
@@ -26,10 +28,9 @@ test.afterEach((done) => {
   done();
 });
 
-test.test('WebSocket connection must connect to server', (test) => {
-  jstp.ws.connect(
+test.test('W3C WebSocket connection must connect to server', (test) => {
+  w3c.connect(
     app.name,
-    null,
     null,
     `ws://localhost:${server.address().port}`,
     (error, conn) => {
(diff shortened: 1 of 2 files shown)


test/node/transport-w3c-ws.js
```diff
@@ -0,0 +1,75 @@
...
+test.afterEach((done) => {
+  if (connection) {
+    connection.close();
+    connection = null;
+  }
+  server.close();
+  done();
+});
+
+test.test('W3C WebSocket connection must connect to server', (test) => {
```
lundibundi: Maybe smth like 'Must connect to the server via W3C WebSocket' and 'Must connect to the server and inspect via W3C WebSocket', here as well as in other places?

test/node/transport-w3c-ws.js
```diff
@@ -0,0 +1,75 @@
...
+  server.listen(0, done);
+});
+
+test.afterEach((done) => {
+  if (connection) {
+    connection.close();
+    connection = null;
+  }
+  server.close();
+  done();
```
belochub: Should you maybe wait for the server to close before calling `done()`?

test/node/transport-w3c-ws.js
```diff
@@ -0,0 +1,75 @@
...
+    interfaces,
+    `ws://localhost:${server.address().port}`,
+    (error, conn, api) => {
+      connection = conn;
+      test.assertNot(error, 'connectAndInspect must not return an error');
+
+      interfaces.forEach((iface) => {
+        test.assert(iface in api, `api must include '${iface}'`);
+        Object.keys(app.interfaces[iface]).forEach(method =>
+          test.assert(method in api[iface],
```
belochub: You should use braces with the lambda in this case.

test/node/transport-w3c-ws.js
nechaido: There is no need, it will get random free port when it starts listening again.

test/node/transport-w3c-ws.js
nechaido: I'll open a follow up PR, that will rename it in every transport test.

test/node/transport-w3c-ws.js
lundibundi: Also AFAIK in all tests we have it this way.

nechaido:
@belochub ping.

belochub:
Landed in https://github.com/metarhia/jstp/commit/6b8a6365ba9890dba73e9ebfd221d343f8860104.

PR #298, opened 2017-08-23
lib, test: add session test

nechaido:


test/utils/session/session-serde.js
```diff
@@ -12,14 +12,13 @@ function stringifySession(session) {
   return JSON.stringify(copy);
 }
 
-function parseSession(sessionString) {
-  const sessionCopy = JSON.parse(sessionString);
-  sessionCopy.buffer = new Map(sessionCopy.buffer);
-  return Object.assign(new jstp.Session(), sessionCopy);
+function deserialize(string) {
```
belochub: I don't like having `string` as an argument name, can we maybe consider changing it to `sessionCopy` or just `session`?

test/node/session-resend-event-on-connection-drop.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const application = new jstp.Application('testApp', ifaces);
+const serverConfig = { applications: [application] };
+const server = jstp.net.createServer(serverConfig);
+
+client.on('message', ([message, ...args]) => {
+  switch (message) {
+    case 'error':
+      test.fail('must not encounter an error');
```
belochub: You probably forgot to also print the error you are passing in the message.

test/node/session-resend-event-on-connection-drop.js
```diff
@@ -0,0 +1,72 @@
...
+});
+
+server.listen(0, () => {
+  client.send(['connect', server.address().port]);
+});
+
+function getClient() {
+  return cp.fork(
+    path.join(__dirname, '../utils/session/client'),
+    [],
```
belochub: You can omit the `args` argument, according to the [docs](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_child_process_fork_modulepath_args_options) it is optional.

test/utils/session/session-serde.js
nechaido: What about `serializedSession`?

test/node/session-resend-event-on-connection-drop.js
```diff
@@ -0,0 +1,72 @@
...
+    { stdin: 'inherit', stdout: 'inherit', stderr: 'inherit' }
```
belochub: There are no such options, and what you probably wanted to do can be done by setting the `silent` option to `false`, which is its default value in Node.js 6 and Node.js 8, meaning you can completely avoid providing options.

test/node/session-resend-event-on-connection-drop.js
nechaido: Fixed.

test/utils/session/session-serde.js
belochub: SGTM

test/node/session-resend-event-on-connection-drop.js
belochub: @nechaido, it hasn't been fixed yet.

nechaido:
Landed in  297c432 and d4710f4.

PR #299, opened 2017-08-28
test: revert tap todo

nechaido:


nechaido:
Landed in 328482e and 02da821.

PR #302, opened 2017-09-07
lint,deps: move ESLint config to a separate package

aqrln:
Move ESLint configuration to a separate package
([`eslint-config-metarhia`](https://github.com/metarhia/eslint-config-metarhia)).  Also, bump ESLint to 4.6.1 (from 4.6.0)
because that's what's specified in that package's peerDependencies.


aqrln:
Failures on Windows due to unsuccessful installation of `uglifyjs-webpack-plugin@0.4.6` seem, obviously, unrelated.

aqrln:
Also, how about converting YAML to JSON in ESLint configs? I've done that in a number of other repos, and JSON looks pretty good with small configs, while being kinda more native than YAML in JS ecosystem.

nechaido:
> Also, how about converting YAML to JSON in ESLint configs? I've done that in a number of other repos, and JSON looks pretty good with small configs, while being kinda more native than YAML in JS ecosystem.

I will not be against it now that the config is small.

aqrln:
@nechaido @lundibundi updated, PTAL

aqrln:
Landed in 79154c9488c29fb37d54b98aae41241a34e098d5.

Issue #217, opened 2017-06-10
Server is crashing on client disconnect after the heartbeat packet


belochub:
This bug occurred with JSTP version 0.5.2.
It is also highly possible that this issue isn't still fixed on master, it needs investigation.
The error log:
```js
TypeError: "listener" argument must be a function;
Connection.once (events.js:307:11);
Connection.setTimeout (/metarhia-jstp/lib/connection.js:258:8);
Connection._heartbeatCallback (/metarhia-jstp/lib/connection.js:208:8);
Connection.startHeartbeat (/metarhia-jstp/lib/connection.js:201:8);
Connection.<anonymous> (/metarhia-jstp/lib/server.js:110:18);
Object.onceWrapper (events.js:293:19); emitNone (events.js:86:13);
Connection.emit (events.js:188:7);
Connection._onTimeout (/metarhia-jstp/lib/connection.js:272:8);
ontimeout (timers.js:386:14);
tryOnTimeout (timers.js:250:5);
Timer.listOnTimeout (timers.js:214:5)
```
@aqrln, may you check this out please?

belochub:
@aqrln, ping.
I believe, that it happens due to the `enroll`/`unenroll` usage for heartbeat timer creation.
Can you investigate this, please?

nechaido:
@aqrln this error resurfaced again on node 8.4.0 on branch `add-distributed-benchmark` while running benchmark.
Steps to reproduce:
1. `node benchmark/distributed/master.js`
2. `node benchmark/distributed/server.js`
3. `node benchmark/distributed/workers -c 1000 -i 50 -w 3`
```

events.js:341
    throw new TypeError('"listener" argument must be a function');
    ^

TypeError: "listener" argument must be a function
    at Connection.once (events.js:341:11)
    at Connection.setTimeout (/home/nechaido/Projects/metarhia/jstp/lib/connection.js:308:10)
    at callback (/home/nechaido/Projects/metarhia/jstp/lib/connection.js:228:12)
    at Connection.startHeartbeat (/home/nechaido/Projects/metarhia/jstp/lib/connection.js:236:5)
    at Connection.connection.on (/home/nechaido/Projects/metarhia/jstp/lib/server.js:105:20)
    at emitTwo (events.js:125:13)
    at Connection.emit (events.js:213:7)
    at Connection._onSessionCreated (/home/nechaido/Projects/metarhia/jstp/lib/connection.js:489:10)
[...]

PR #303, opened 2017-11-22
connection: use ping-pong instead of heartbeat

nechaido:
* Send Ping-Pong instead of Heartbeat messages.
* Use setInterval() instead of timers.enroll().
* Rewrite tests.

Fixes: #217

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -236,7 +236,7 @@ class Connection extends EventEmitter {
 
   // Send a pong message
   //
-  pong(messageId) {
+  _pong(messageId) {
     if (this._closed) {
       return;
     }
@@ -252,10 +252,7 @@ class Connection extends EventEmitter {
       if (this._closed) {
         return;
       }
-      const pingId = this._nextMessageId;
-      this.ping(() => {
-        this.session.unbufferUpTo(pingId);
-      });
+      this.ping();
     };
 
     this._heartbeatInterval = setInterval(heartbeat, interval);
@@ -683,7 +680,7 @@ class Connection extends EventEmitter {
   //
   _processPingMessage(message) {
     const messageId = message.ping[0];
-    this.pong(messageId);
+    this._pong(messageId);
     this.session._onMessageRecieved(messageId);
   }
 
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -52,7 +52,7 @@ class Connection extends EventEmitter {
     this.application = null;
     this.remoteProxies = {};
 
-    this._heartbeatInterval = null;
+    this._heartbeatTimer = null;
     this._closed = false;
 
     // Defined in constructor to be used as default callback in callMethod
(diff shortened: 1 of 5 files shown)


lib/connection.js
```diff
@@ -241,33 +237,36 @@ class Connection extends EventEmitter {
   // Send a pong message
   //
   pong(messageId) {
+    if (this._closed) {
```
aqrln: Given that it's a public method, I'd say that the caller shouldn't invoke `pong()` if the connection is closed. Otherwise, this check makes `pong()` inconsistent with other methods of a connection in a way that one may call it on a closed connection without any errors, which doesn't sound right.

nechaido: @aqrln mark `pong` as a private method, there is no reason for it to be public.

aqrln: @nechaido sounds good to me, but the same goes for `callback`, then. (Though it's kinda out of scope of this PR.)

lib/connection.js
```diff
@@ -53,18 +52,15 @@ class Connection extends EventEmitter {
     this.application = null;
     this.remoteProxies = {};
 
-    this._heartbeatCallbackInstance = null;
+    this._heartbeatInterval = null;
```
belochub: I think it will be more appropriate to call this field `_heartbeatTimer` or `_heartbeatTimeout` because it is somewhat confusing to use `interval` here, due to the same word being used as a parameter name in `startHeartbeat` method and having the meaning of the delay between heartbeats.

lib/connection.js
```diff
@@ -240,34 +236,34 @@ class Connection extends EventEmitter {
...
   //
   startHeartbeat(interval) {
-    const callback = () => {
-      this.transport.send('{}');
-      this.setTimeout(interval, this._heartbeatCallbackInstance);
-
-      if (process.env.NODE_ENV !== 'production') {
-        this.emit('heartbeat', this._heartbeatMessage);
+    const heartbeat = () => {
+      if (this._closed) {
```
belochub: This can be simplified to
```javascript
if (!this._closed) {
  this.ping();
}
```

lib/connection.js
```diff
@@ -400,6 +374,7 @@ class Connection extends EventEmitter {
   // Closed socket event handler
   //
   _onSocketClose() {
+    this.closed = true;
```
belochub: Looks like there is a typo here, it should be `this._closed` instead of `this.closed`.

belochub:
There is no need to add `test` label to this PR, it is similar to the situation in https://github.com/metarhia/jstp/pull/190#issuecomment-304742538.

aqrln:
> There is no need to add test label to this PR, it is similar to the situation in #190 (comment).

Yup, not even `lib`. The subsystem should be just `connection`.

nechaido:
@belochub @aqrln ping.

nechaido:
Landed in 37918fec691890572cc21ddd0d92fa514bc5800a.

PR #306, opened 2017-11-22
connection: make callback method private

aqrln:
There's no reason to call it directly outside of the connection's
internal machinery except to break things.

Also, this will make it consistent with `pong` after #303 lands.


nechaido:
Landed in 18a6ef97f4982b505bf768fa4876bbbad440bfd0.

PR #307, opened 2017-11-22
test: update Travis config to use the current Node

aqrln:
Remove end-of-life Node.js 7 and add the current Node.js 9.


nechaido:
Landed in 6cb4762.

PR #309, opened 2017-12-05
server: add a way to update API

nechaido:


lib/server.js
```diff
@@ -74,6 +79,10 @@ class Server {
     return Array.from(this.getClients());
   }
 
+  updateApi(applications) {
```
belochub: I think you should add a comment for this function.

nechaido:
@lundibundi

lundibundi:
> This functionality is for mainly for adding new applications or versions. Moreover, you should never change existing version's API, add a new version instead.

@nechaido Well that's why a separate method to add new versions or applications is preferable to the one that requires you to pass all existing ones. The one I mentioned will allow you to do just that.

nechaido:
@lundibundi please feel free to add described functionality in a follow-up PR.
I would prefer to keep this PR simple so that we land it ASAP because the implemented functionality is critically needed to integrate JSTP with Impress.

lundibundi:
@nechaido Oh, that's the reason behind this PR. Then it's okay with me.

lib/server.js
```diff
@@ -22,26 +40,16 @@ const HANDSHAKE_TIMEOUT = 3000;
...
-  this._cachedVersions = new Map();
-  applications.forEach((appVersions, appName) => {
-    const versions = Array.from(appVersions)
-      .filter(version => version[0] !== 'latest')
-      .map(version => [new semver.SemVer(version[0]), version[1]])
-      .sort((a, b) => semver.rcompare(a[0], b[0]));
-    this._cachedVersions.set(appName, versions);
-  });
+  ({
+    cachedVersions: this._cachedVersions,
```
aqrln: That's wizardry.

lib/server.js
```diff
@@ -8,6 +8,24 @@ const SimpleAuthPolicy = require('./simple-auth-policy');
 
 const HANDSHAKE_TIMEOUT = 3000;
 
+const _initApplications = function(applications) {
```
aqrln: This function doesn't need to start with an underscore, it isn't exported anyway.

nechaido:
Landed in d5cb305e31abd995351b59de22cecc22f3212e6b.

PR #310, opened 2017-12-06
server: add a way to move connections to a new app version

nechaido:


lib/applications.js
```diff
@@ -85,41 +85,47 @@ class Application {
...
+      appIndex.set(application.name, versions);
+    }
+    versions.push(application);
+  });
+
+  appIndex.forEach((instances, name) => {
+    const versions = new Map();
+    index.set(name, versions);
+    if (instances.length === 1) {
+      versions.set('1.0.0', instances[0]);
```
lundibundi: You should check if app has version and set 1.0.0 only if it doesn't. You discard app version here.

lib/connection.js
```diff
@@ -447,16 +447,20 @@ class Connection extends EventEmitter {
...
+      applicationName, this.requestedVersion
     );
 
     if (!application) {
       this._handshakeError(errors.ERR_APP_NOT_FOUND);
       return;
     }
 
+    if (!this.requestedVersion) {
+      this.requestedVersion = semver.major(application.version).toString();
```
lundibundi: In case when there is only one app without version and no version specified upon handshake call this still seems to call semver.major on undefined, or have I missed something?

nechaido: Fixed.

lib/server.js
```diff
@@ -92,6 +92,19 @@ class Server {
     } = prepareApplications(applications));
   }
 
+  // Update applivation version for each connection.
+  updateConnections() {
+    for (const connection of this.clients.values()) {
+      if (semver.valid(connection.requestedVersion)) {
```
lundibundi: It's better to add a comment that this function will not update clients who specified direct versions and not version-range upon request in order to make this clearer. Or just comment here like 'ignore clients who requested specific API versions'.

test/node/api-versions.js
```diff
@@ -228,3 +217,34 @@ test.test('must throw an error on invalid version in application parameter',
...
+      test.assertNot(error, 'connect must not return an error');
+      connection.callMethod('calculator', 'answer', [], (error, result) => {
+        test.assertNot(error, 'callMethod must not return an error');
+        test.strictSame(result, 13);
+        test.end();
+      });
+    });
+  });
+});
+
```
lundibundi: I think we should also add test that'll test 'must connect to app without version('latest') when client didn't specify version'

test/node/update-api.js
```diff
@@ -55,3 +90,79 @@ test.test('must update API', (test) => {
     }
   );
 });
+
+test.test(
+  'must not update API on existing connection if no supported version is found',
```
lundibundi: We should clarify test description here and in the tests after this one (probably adding 'version' should be enough), it'll be hard to understand the problem if one of those fails.

test/node/api-versions.js
nechaido: @lundibundi there is no app without version now, and we already have a test that checks if a user connects to the latest app, if no version was specified.

lib/server.js
```diff
@@ -92,6 +92,19 @@ class Server {
     } = prepareApplications(applications));
   }
 
+  // Update applivation version for each connection.
+  updateConnections() {
```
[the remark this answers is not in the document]
nechaido: I don't like both of them, maybe someone has better ideas @metarhia/jstp-core ?

belochub:
Rebase this on master and rename it to correspond to our general naming style, please.

test/node/update-api.js
nechaido: @lundibundi could you please clarify proposed change?

lundibundi: @nechaido We can add version to the description ''must not update API on existing connection if no supported version is found (connect to @1)" and so with other tests in order to clearly discern them should an error occur.

lib/server.js
```diff
@@ -92,6 +92,22 @@ class Server {
     } = prepareApplications(applications));
   }
 
+  // Update application version for each connection.
+  // It will update to a newest version in the requested range. If client
+  // requested version in (1.1) range he will not be updated to (1.2.x). If no
```
belochub: I think it'll be better to specify the first range as `(1.1.x)` instead of just `(1.1)` for consistency with the second one.
You should also add a comma after 'range': `'If client requested version in (1.1.x) range, he will...'`.

lib/server.js
```diff
@@ -92,6 +92,22 @@ class Server {
...
+  // suitable version is found client will use it's previous app version.
```
belochub: Missing a comma and an article here: `'...is found, the client will...'`.
Also, `'it's'` shouldn't have an apostrophe, it is a possessive here.

lib/server.js
```diff
@@ -92,6 +92,22 @@ class Server {
...
+  updateConnectionsApi() {
+    for (const connection of this.clients.values()) {
+      if (semver.valid(connection.requestedVersion)) {
+        continue;
+      }
+      const newApp = this._getApplication(
+        connection.application.name, connection.requestedVersion
+      );
+      connection.application = newApp || connection.application;
```
belochub: In my opinion, it is better to use `if` here for readability, because in the case when `!newApp` is `true` it comes down to the assignment of the variable to itself, which isn't more clear than just doing
```javascript
if (newApp) {
  connection.application = newApp;
}
```

test/node/api-versions.js
```diff
@@ -58,9 +59,15 @@ test.test('must allow to specify version in application name', (test) => {
   test.end();
 });
 
+test.test('must set version to (1.0.0) if none was provided', (test) => {
```
belochub: Why does this test have the same description as the one on line 221 in this file, do they test the same thing?

nechaido: I forgot to delete a test on line 221. It is deleted now.

nechaido:
Landed in f94b50e and 23b8b18.

PR #312, opened 2018-01-25
connection: emit logging info from connection on a server

nechaido:


lib/connection.js
```diff
@@ -717,6 +717,13 @@ class Connection extends EventEmitter {
   _remoteCallbackWrapper(messageId, error, ...result) {
     this._callback(messageId, error, result);
   }
+
+  _log(message, ...args) {
```
belochub: It's not actually a `message`, I think, it should be called `eventName`.

lundibundi: I agree, but maybe smth like `messageType` will be better.

lib/connection.js
```diff
@@ -717,6 +717,13 @@ class Connection extends EventEmitter {
...
+    this.emit(message, ...args);
+    if (this.server) {
+      this.server.emit(message, ...args, this);
```
lundibundi: I think connection argument should be second and not the last, as this will ease using a general listener that'll get a connection as a second parameter and not extract it from the array of all args.

nechaido:
@lundibundi ping.

lib/connection.js
```diff
@@ -717,6 +717,13 @@ class Connection extends EventEmitter {
   _remoteCallbackWrapper(messageId, error, ...result) {
     this._callback(messageId, error, result);
   }
+
+  _emitEventForLogging(event, ...args) {
+    this.emit(event, ...args);
+    if (this.server) {
+      this.server.emit('log', this, event, ...args);
```
lundibundi: Oh, this is even better, IMHO.

belochub:
Landed in https://github.com/metarhia/jstp/commit/eb4d54c47696a36aeee72a0aa14a2d2763cb52ca.

PR #315, opened 2018-01-29
connection: reject invalid event message

belochub:
Reject event messages having anything other than array provided as event arguments.


belochub:
Landed in https://github.com/metarhia/jstp/commit/2f7db7947c36f8b6eb0282a3d7c54a0862613d22.

PR #317, opened 2018-01-29
server: fix comment explaining authPolicy argument

belochub:


lib/server.js
```diff
@@ -34,8 +34,15 @@ const prepareApplications = function(applications) {
...
-//                see jstp.SimpleAuthPolicy.
+//                authenticate (optional).
+//                This method takes next arguments:
+//                  connection - jstp.Connection object.
+//                  application - application object.
+//                  strategy - authentication strategy string (note: this method
+//                             will NOT be called for 'anonymous' and 'session'
+//                             authentication strategies).
+//                  credentials - authentication credentials array.
+//                  callback - callback with signature (error, username).
```
lundibundi: In `connection.js` we use 'callback - callback function ...' [see](https://github.com/metarhia/jstp/blob/95b1d66f410cc8c2426a1bda20afdd4899c01bd8/lib/connection.js#L73) maybe we should write it that way here too, to be consistent. Even though I think it's better for it to be 'callback - function ...'

belochub:
Landed in https://github.com/metarhia/jstp/commit/efd8d8d18d781cc300fc57a6d17bd85230489556.

PR #319, opened 2018-01-31
connection: use Map for storing callbacks

belochub:


belochub:
Landed in https://github.com/metarhia/jstp/commit/a6c60cfa78fa67eab2a854f3ef131f7850c59960.

PR #320, opened 2018-01-31
lib: add utility for call messages resending

belochub:
Callbacks passed to jstp method calls will now return an error when the
connection is being closed.
Since it is known, that callback messages are not being resent on connection
drop, a special method on Connection was added that resends call message on
connection to the server, and thus can be used for making a call to idempotent
methods on the server.

I would like to request some help with tests from @nechaido.
Also, there is a question about how we should deal with calls to remote methods done via RemoteProxy, I don't think that we should definitely resend all of the calls, but we should probably provide some possibility for the user to indicate that the call must be resent. Does anyone have any idea on how this should be done?

### DIFF
--- a/lib/remote-proxy.js
+++ b/lib/remote-proxy.js
@@ -19,17 +19,24 @@ class RemoteProxy extends EventEmitter {
     methods.filter(method => this[method] === undefined)
       .forEach((method) => {
         this[method] = (...args) => {
-          let callback = args[args.length - 1];
+          let resend = args[args.length - 1];
+          let callback = args[args.length - 2];
 
-          if (typeof(callback) === 'function') {
-            args = Array.prototype.slice.call(args, 0, -1);
+          if (typeof callback === 'function' && typeof resend === 'boolean') {
+            args = args.slice(0, -2);
+          } else if (typeof resend === 'function') {
+            callback = resend;
+            resend = false;
+            args = args.slice(0, -1);
           } else {
             callback = null;
+            resend = false;
           }
 
-          this._connection.callMethod(
-            this._interfaceName, method, args, callback
-          );
+          (resend ?
+            this._connection.callMethodWithResend :
+            this._connection.callMethod
+          ).call(this._connection, this._interfaceName, method, args, callback);
         };
       });
   }
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -123,6 +123,10 @@ class Connection extends EventEmitter {
     return removedTransport;
   }
 
+  _getCallbackId(absId) {
+    return absId * this._messageIdDelta;
+  }
+
   // Send a call message over the connection
   //   interfaceName - name of an interface
   //   methodName - name of a method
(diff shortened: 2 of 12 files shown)


belochub:
Ping @lundibundi.

belochub:
Ping @lundibundi.

lundibundi:
@belochub @nechaido I do not see a reason behind all of the complexity of `resendable-call-from-client.js` test. As I understand it's supposed to test resending of calls that were sent at a point where client assumed that connection was still active but the client did not get the reply for one reason or another. So why not just make this test the same as  `resendable-call-from-closed-client.js` but instead of calling [connection.close()](https://github.com/metarhia/jstp/pull/320/files#diff-db589711a99d920199d016b04beeae15R27) close the server-side connection there via `server.getClientsArray()[0].close();`, this way at a point of sending a call the client still has an active connection (from his point of view) and the server will not receive this call?

lundibundi:
Also, I've been thinking about resending calls that have not been received by the other party, won't we make the same call twice this way?
So the example is as follows if we make a `callMethodWithResend` and after it, the connection breaks so that the other party didn't receive the call but it was buffered to be sent again on this side. As this side's connection was closed `_dismantleCallbacks` was called and another call was buffered. Hence upon reconnecting the session will resend both the original call (because the other party notified that it didn't receive it by sending appropriate receivedMessagesCount) and the one after it but as the original's call callback was removed only one callback will be called, but we still have 2 calls.
I have slightly modified resendable-call-from-closed-client.js test by replacing `connection.close()` with server-side socket destroy and was able to witness the problem I've described above.

nechaido:
> @belochub @nechaido I do not see a reason behind all of the complexity of resendable-call-from-client.js test. As I understand it's supposed to test resending of calls that were sent at a point where client assumed that connection was still active but the client did not get the reply for one reason or another. So why not just make this test the same as resendable-call-from-closed-client.js but instead of calling connection.close() close the server-side connection there via server.getClientsArray()[0].close();, this way at a point of sending a call the client still has an active connection (from his point of view) and the server will not receive this call?

Thank you, I have added the corresponding test.

> Also, I've been thinking about resending calls that have not been received by the other party, won't we make the same call twice this way?
So the example is as follows if we make a callMethodWithResend and after it, the connection breaks so that the other party didn't receive the call but it was buffered to be sent again on this side. As this side's connection was closed _dismantleCallbacks was called and another call was buffered. Hence upon reconnecting the session will resend both the original call (because the other party notified that it didn't receive it by sending appropriate receivedMessagesCount) and the one after it but as the original's call callback was removed only one callback will be called, but we still have 2 calls.
I have slightly modified resendable-call-from-closed-client.js test by replacing connection.close() with server-side socket destroy and was able to witness the problem I've described above.

Fixed. It was the problem with the test itself.

lundibundi:
@belochub basically the issue is with `_dismantleCallbacks()` and `callMethodWithResend`. As `_dismantleCallbacks` is currently called upon connection close, but at that time you don't yet know whether the user will not get a callback or he will (because the message might not have been send at all hence it will be resent as usual, see [comment](https://github.com/metarhia/jstp/pull/320#issuecomment-366461027)) hence it is incorrect to schedule resend of messages upon connection close. You will only know whether user will or will not get a callback when he/server sends back a number of messages it received. That's why in  java [implementation](https://github.com/metarhia/jstp-java/commit/e90ab2a1ec67cff291302ce47296077bdcd49f00#diff-8380b5e9c1da2badfff36ff5ff58a320R51) the error is actually reported upon connection restore.

nechaido:
@lundibundi ping.

lib/remote-proxy.js
```diff
@@ -19,17 +19,31 @@ class RemoteProxy extends EventEmitter {
...
+            this._connection.callMethodWithResend(
+              this._interfaceName, method, args, callback
+            );
+            return;
           }
 
           this._connection.callMethod(
             this._interfaceName, method, args, callback
           );
+
```
lundibundi: Is this blank line needed?
Also, I'd suggest to just use if\else on the line above as it should look better imo.

belochub: @lundibundi, if/else will increase indentation in the second block, I don't think that will look better.

lundibundi: I meant conceptually it will present the actual if/else (as that's what's happening here) and avoid a return hack for indentation purposes only. I'd understand if it affected the whole/almost whole body of a function but here it's just one statement.

belochub:
Landed in https://github.com/metarhia/jstp/commit/19402edc4863b7fddde5f538eda56a4ad0f59c21.

PR #65, opened 2017-02-10
build: fail CI if native addon build fails

aqrln:
Fixes: https://github.com/metarhia/JSTP/issues/63

Issue #63, opened 2017-02-09
Build should fail when native extension can't be built on CI


belochub:
I checked CI logs today and found out that our native extension was not built for a while now because of compilation errors. It caused no errors in CI, because it just used JavaScript implementation of parser which is working fine.
I think that we maybe need to add some environment variable check for CI so that tests won't pass when native extension can't be built.

aqrln:
@belochub yeah, that's quite easy to check. The variable is `process.env.TRAVIS`.

PR #325, opened 2018-02-08
build: fail CI if native addon build fails

nechaido:
This is an additional patch to that made in https://github.com/metarhia/jstp/pull/65 that enables failing for CI systems other than Travis.

Refs: https://github.com/metarhia/jstp/issues/63


belochub:
Landed in https://github.com/metarhia/jstp/commit/36c06363e41f5e087464bdb612e3c51136f1ad2c.

PR #329, opened 2018-02-28
lib: add missing callback call

belochub:
Fix callback not being called in `connectAndInspect` family of functions when an empty array was provided in `interfaces` argument.


belochub:
Landed in https://github.com/metarhia/jstp/commit/0ae0c1c681c2db5fd78abec9d080eade2e7226c2.

PR #331, opened 2018-03-06
deps: update dependencies

belochub:
Also switch to `babel-preset-env` instead of `babel-preset-es2015`, since the latter one is deprecated (http://babeljs.io/env).


belochub:
Landed in https://github.com/metarhia/jstp/commit/87e330a5f79648d94a8e796c873357fe2cf0529d.

PR #332, opened 2018-04-26
lib: realize reconnection by transport replacement

belochub:
This is also to be covered by unit tests, I hope @nechaido can help with that.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const crypto = require('crypto');
+
 // Forward an event from one EventEmitter to another.
 //   from - EventEmitter to listen for event
 //   to - EventEmitter to emit event on
@@ -66,6 +68,75 @@ const rsplit = (string, separator) => {
   return [string.slice(0, lastIndex), string.slice(lastIndex + 1)];
 };
 
+// Value prefetcher to use when crypto.randomBytes is required to generate
+// multiple same-size values. `bufSize` must be a multiple of `valueSize` for
+// it to work.
+//
+class CryptoRandomPrefetcher {
+  constructor(bufSize, valueSize) {
+    this.buf = crypto.randomBytes(bufSize);
+    this.pos = 0;
+    this.vsz = valueSize;
+  }
+
+  next(
+    // Returns Buffer with next `valueSize` random bytes.
+  ) {
+    if (this.pos === this.buf.length) {
+      this.pos = 0;
+      crypto.randomFillSync(this.buf);
+    }
+    const end = this.pos + this.vsz;
+    const buf = this.buf.slice(this.pos, end);
+    this.pos = end;
+    return buf;
+  }
+}
+
+const randPrefetcher = new CryptoRandomPrefetcher(4096, 4);
+const UINT32_MAX = 0xFFFFFFFF;
+
+// Generate random number in the range from 0 inclusive up to
+// but not including 1 (same as Math.random),
+// but use crypto-secure number generator.
+//
+const cryptoRandom = () =>
+  randPrefetcher.next().readUInt32LE(0, true) / (UINT32_MAX + 1);
+
+// Special purpose map, that does not immediately delete the values when
+// `delete` method is being called, but postpones the deletion instead, every
+// time the `delete` method is called, to some moment in the future, but not
+// later than two times the `interval`. Also, whenever `get` is called on the
+// key that is awaiting its deletion, the deletion is aborted.
+//
+class ExpiringMap extends Map {
+  constructor(interval, ...args) {
+    super(...args);
+
+    this._interval = interval;
+    this._timer = setInterval(() => this._purge(), interval).unref();
+    this._deletionTimes = new Map();
+  }
+
+  _purge() {
+    const purgeTime = Date.now();
+    this._deletionTimes.forEach((delTime, key) => {
+      if (purgeTime - delTime > this._interval) {
+        super.delete(key);
+      }
+    });
+  }
+
+  get(key) {
+    this._deletionTimes.delete(key);
+    return super.get(key);
+  }
+
+  delete(key) {
+    this._deletionTimes.set(key, Date.now());
+  }
+}
+
 module.exports = {
   forwardEvent,
   forwardMultipleEvents,
@@ -74,4 +145,6 @@ module.exports = {
   extractCallback,
   doNothing,
   rsplit,
+  cryptoRandom,
+  ExpiringMap,
 };
--- a/lib/server.js
+++ b/lib/server.js
@@ -5,6 +5,7 @@ const semver = require('semver');
 const apps = require('./applications');
 const Connection = require('./connection');
 const errors = require('./errors');
+const { ExpiringMap } = require('./common');
 const Session = require('./session');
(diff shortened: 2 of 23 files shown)


tshemsedinov:
@belochub I think `implement reconnection` is the right caption

belochub:
@tshemsedinov in this context, the words 'to realize' and 'to implement' are equivalent, and I've picked 'to realize' since it is shorter and using this word was the only way to fit the commit summary in 50 characters.

belochub:
@lundibundi, ping, can you, please, take a look at this PR?

belochub:
Ping, @lundibundi.

lundibundi:
@belochub Hi, sorry, I'm a bit busy right now and don't have time to properly review this PR, feel free to merge this without my review.

lib/server.js
```diff
@@ -204,11 +215,26 @@ class Server {
...
+      connection.handshakeDone = true;
+      connection = existingConnection;
+      const cb = connection._onSessionRestored.bind(connection);
+      getSessionCallback(connection.session, replacementTransport, cb);
+      return;
+    }
+
+    if (sessionStorageProvider.isAsync()) {
+      sessionStorageProvider.get(sessionId, null, session =>
+        getSessionCallback(session, callback)
```
lundibundi: It seems you have missed the second argument here (`session, null, callback`)?

lib/server.js
```diff
@@ -204,11 +215,26 @@ class Server {
...
-      getSessionCallback(existingConnection.session);
-    } else if (sessionStorageProvider.isAsync()) {
-      sessionStorageProvider.get(sessionId, getSessionCallback);
+      const replacementTransport = connection.transport;
+      Object.keys(connection._transportListeners).forEach((event) => {
+        connection.transport.removeListener(
+          event, connection._transportListeners[event]
+        );
+      });
+      connection.transport = null;
```
lundibundi: I think it's better to move this (218-224) to Connection to something like `_removeTransport()` that will return removed transport?

lib/transport-common.js
```diff
@@ -5,6 +5,30 @@ const Application = require('./applications').Application;
...
+const MAX_BACKOFF_TIME = 32000;
+const RANDOM_OFFSET_MAX = 1000;
+
+function createDefaultReconnector() {
+  let backoff = DEFAULT_BACKOFF_TIME;
+  const randomMsecGen = () =>
+    Math.floor((RANDOM_OFFSET_MAX + 1) * common.cryptoRandom());
+  return (connection, reconnectFn) => {
+    if (connection.closedIntentionally) return;
+    setTimeout(() => {
```
lundibundi: Do I understand it correctly that default reconnector will try to reconnect indefinitely after reaching `MAX_BACKOFF_TIME` with `MAX_BACKOFF_TIME` in case there is always an error or have I missed something? If it's the case I think it'd be better to stop after trying with `MAX_BACKOFF_TIME` once or at least make amount of retries configurable.

lib/common.js
```diff
@@ -66,6 +68,60 @@ const rsplit = (string, separator) => {
...
+    });
+  }
+
+  get(key) {
+    this._deletionTimes.delete(key);
+    return super.get(key);
+  }
+
+  delete(key) {
+    this._deletionTimes.set(key, Date.now());
```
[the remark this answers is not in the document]
belochub: This class was created for internal usage, where it is not supposed to call `delete` on the same key multiple times in a row without calling `get` in between those calls to make sure, that the required item is still there, thus resetting the deletion time.

lib/server.js
belochub: Yeah, thanks.

lib/transport-common.js
belochub: You can call `connection.close` to stop reconnection with this default reconnector.
Also, users of the library can always provide their own reconnector implementation that may behave like you described or that may have any other different logic behind it.
I don't see how we can make the number of retries configurable (since it is the default reconnector, used when none is provided) and I don't think that it is okay to just let us make process termination (when reconnection is the only thing keeping it alive) default behavior if the client isn't able to connect successfully during the period of ~70 seconds.

lib/server.js
belochub: Yeah, sure.

lib/common.js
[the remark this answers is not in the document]
lundibundi: Well, even if it's internal we still have to know how it works and this is clearly not an obvious behavior. In the end it's in lib/common which means it may be used later for other things and lead to strange behavior.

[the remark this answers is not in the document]
belochub: It is expected to work like that by the code that is using it internally and it is covered by tests, so I can't say that this behavior is not obvious. Also, it is not exported outside, and I don't think it may be used anywhere else since it's usage is very specific in its nature.
Regarding the placement of this class in this file, do you think it will be better to place this class in `lib/server.js` instead?

lib/transport-common.js
lundibundi: > I don't see how we can make the number of retries configurable

Oh, I thought that we expose this function to the user, that's why I suggested it. I think we should expose it as backoff is a common strategy. This way we can just add an argument that will control this behavior.

>  I don't think that it is okay to just let us make process termination default behavior

But this process will only use user resources without any utility to the user and the only way to stop it will be `kill`. I don't think it's appropriate to keep alive useless process without explicit approval of the user. Which also adds to the point of just exposing this function and giving user the control. As for the default behavior I think it's better to just stop trying but maybe @nechaido  can add smth?

belochub: >But this process will only use user resources without any utility to the user and the only way to stop it will be `kill`.

I mean, retries are asynchronous, `kill` is not the only option to terminate the process.
A user can always use `connection.close` to stop reconnection, this was initially @nechaido's idea.

>to keep alive useless process without explicit approval of the user

What do you mean by 'useless'? It is not useless, if the process is supposed to communicate with the server after the connection is established, then it is still the task of this process to keep the connection alive and not just terminate the process when the connection is lost for a minute, possibly losing the data. A client can even make some calls when the connection is lost, and they will be transmitted to the server when it reconnects successfully, and if the client doesn't want it to happen, it can use `connection.close` method explicitly closing the connection.

I am okay with exposing the function, but in that case, we need to decide on what things will be configurable and we should also create some other alternatives, like the one disabling reconnection, for example (possibly making it the default one, to avoid changing the behavior). Also, I am not sure about how it should behave with this argument specifying the number of retries, is it a total count of retries since the first, or is it the count of retries since the first time maximum backoff time was used?

nechaido: I am okay with exposing the function, in case we make it configurable, but I would prefer that its default behaviour stays the same.

lib/common.js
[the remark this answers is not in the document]
belochub: @lundibundi I've provided documentation to all of the new methods and classes in this file, including the `ExpiringMap`, it should be clear now, take a look.

belochub:
Landed in https://github.com/metarhia/jstp/commit/8ff6423aae87ac222c98b6bdba6a8477f7aa6a2c and https://github.com/metarhia/jstp/commit/e1d8a920ecd7f73abca9250fa1ba41e684b10f18.

PR #339, opened 2018-06-07
connection: fix incorrect client-side message IDs

belochub:
Fix incorrect client-side message IDs, occurring after reconnection to
the existing session.


belochub:
Landed in https://github.com/metarhia/jstp/commit/644b7d55971f1c278917669362bd19032679a5ce.

PR #342, opened 2018-06-26
server: enable returning errors from authPolicy

igorgo:
It would be very useful to return real authentication errors to the client, such as `Account locked`, `Password expired`, etc.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -166,7 +166,17 @@ class Server {
     this.authenticate(
       connection, application, strategy, credentials,
       (error, username) => {
-        if (error || !username) {
+        if (error) {
+          if (error instanceof errors.RemoteError) {
+            callback(error);
+          } else {
+            callback(new errors.RemoteError(
+              errors.ERR_AUTH_FAILED, error.message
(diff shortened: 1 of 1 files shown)


lib/server.js
```diff
@@ -166,7 +166,17 @@ class Server {
     this.authenticate(
       connection, application, strategy, credentials,
       (error, username) => {
-        if (error || !username) {
+        if (error) {
+          if (error instanceof errors.RemoteError) {
+            callback(error);
+          } else {
+            callback(error.message ?
```
nechaido: it's better to write it this way: `callback(new RemoteError(errors.ERR_AUTH_FAILED, error.message))`.

igorgo: Оh, sorry. The fragment in your comment is not full and I`ve misunderstood.  D'u mean to use your code instead of ternary only? Then yes,  I agree.

belochub:
Landed in https://github.com/metarhia/jstp/commit/f108ab627cb31b2ed645f1ccd7d01b30a65e3e5e.

PR #343, opened 2018-07-02
meta: update AUTHORS

belochub:


tshemsedinov:
@belochub CI built only after I restarted it 🤷

PR #346, opened 2018-07-13
server: fix call to async sessionStorageProvider

belochub:
Fix incorrect usage of asynchronous sessionStorageProviders by the
server.


belochub:
Landed in https://github.com/metarhia/jstp/commit/f816d209a2440d9115e0f6edecf62b46b0b005a1.

PR #353, opened 2018-07-31
deps,lint: update eslint config

nechaido:
Also fix all linter errors produced by `implicit-arrow-linebreak` rule.


lundibundi:
What was the reason behind `implicit-arrow-linebreak` rule? I see no harm in having 2 styles of arrow functions, furthermore, I personally prefer to have it without parens or just with curly brackets. Can we remove it from our config?

nechaido:
@lundibundi we can override the rule in our config, but I would prefer to have the same code style throughout the whole organisation, thus the whole idea of `eslint-config-metarhia`.

lundibundi:
@nechaido well, I thought to override it in `eslint-config-metarhia`, that's why I /cc'd @aqrln as he was the one who added it there. I just wanted to discuss it here before creating an issue/changing it there as I have missed that change when it was committed.

belochub:
@lundibundi, there is a way to not use parentheses with this rule, we just have to make sure that function body begins on the same line as it's description, so that '`=>`' is not immediately followed by a line break.

aqrln:
I added the rule when it became available in ESLint because that had already been the house style throughout the org, with JSTP, probably, being the only exception, and I'm pretty sure @tshemsedinov was in favor of this style strongly.

aqrln:
That said, I personally have no strong opinion one way or the other, and consistency within the org was the sole motivation. Today I'd rather just use Prettier, tbh.

test/node/connection-inspect.js
```diff
@@ -26,8 +26,11 @@ test.afterEach((done) => {
 
 test.test('must process inspect messages', (test) => {
   const expectedInterfaces = Object.keys(app.interfaces);
-  const expectedTests = expectedInterfaces.reduce((tests, iface) =>
-    tests + Object.keys(app.interfaces[iface]).length + 1, 1);
+  const expectedTests = expectedInterfaces.reduce(
+    (tests, iface) => (
```
aqrln: and to this one

test/node/connection-inspect.js
```diff
@@ -26,8 +26,11 @@ test.afterEach((done) => {
...
+      tests + Object.keys(app.interfaces[iface]).length + 1
+    ), 1
```
aqrln: Nit: could you please also move the second argument to a new line?

test/node/connection-inspect.js
```diff
@@ -38,9 +41,9 @@ test.test('must process inspect messages', (test) => {
     expectedInterfaces.forEach((iface) => {
       connection.inspectInterface(iface, (error, methods) => {
         test.assertNot(error, `must inspect ${iface}`);
-        Object.keys(app.interfaces[iface]).forEach(method =>
+        Object.keys(app.interfaces[iface]).forEach(method => (
```
aqrln: Ditto, I agree with the linter on this one. Either this, or the function itself should be moved to a new line if we allow line-breaking arrow functions without parens. I find the left-hand side of the diff a bit weird to read.

test/node/connection-inspect.js
```diff
@@ -57,10 +60,10 @@ test.test('must generate remote api after inspect', (test) => {
 
       expectedInterfaces.forEach((iface) => {
         test.assert(iface in api, `api must include '${iface}'`);
-        Object.keys(app.interfaces[iface]).forEach(method =>
+        Object.keys(app.interfaces[iface]).forEach(method => (
```
aqrln: Ditto there and everywhere else this pattern might occur.

lib/server.js
```diff
@@ -234,9 +234,9 @@ class Server {
     }
 
     if (sessionStorageProvider.isAsync()) {
-      sessionStorageProvider.get(sessionId, session =>
+      sessionStorageProvider.get(sessionId, session => (
```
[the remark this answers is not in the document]
lundibundi: imo it's even better to have it in just 2 lines (no additional pattern and put the closing pattern on the same line as body)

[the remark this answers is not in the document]
aqrln: that works for me too, yes

nechaido:
@lundibundi @aqrln PTAL.

lib/socket.js
```diff
@@ -109,28 +109,29 @@ const socketFactory = (connect) => {
...
+    );
 
 // Same as newConnectFn but will instead perform reconnect on existing
 // connection.
 //
 // see transportCommon.newReconnectFn
 //
-const newReconnectFn = connect =>
-  transportCommon.newReconnectFn(socketFactory(connect), Transport);
+const newReconnectFn =
```
belochub: You have enough space left to start function definition on the same line as you did on line 112 in this file.

lib/ws-browser.js
```diff
@@ -123,10 +123,13 @@ const reconnect =
...
-    connectAndInspect(app, client, interfaces, url, callback),
-  reconnect: (connection, url, callback) =>
-    reconnect(connection, url, callback),
+  connect: (app, client, url, callback) => connect(app, client, url, callback),
+  connectAndInspect:
+    (app, client, interfaces, url, callback) => connectAndInspect(
+      app, client, interfaces, url, callback
+    ),
+  reconnect:
+    (connection, url, callback) => reconnect(
```
belochub: You can fit complete call to `reconnect` on this line.

lib/tls.js
```diff
@@ -65,12 +65,11 @@ const reconnect = sock.newReconnectFn(tls.connect);
...
-  connect: (app, client, ...options) =>
-    connect(app, client, ...options),
-  connectAndInspect: (app, client, interfaces, ...options) =>
-    connectAndInspect(app, client, interfaces, ...options),
-  reconnect: (connection, ...options) =>
-    reconnect(connection, ...options),
+  createServer: (options, listener) => createServer(options, listener),
+  connect: (app, client, ...options) => connect(app, client, ...options),
+  connectAndInspect:
+  (app, client, interfaces, ...options) => connectAndInspect(
```
belochub: Shouldn't indentation increase here, like everywhere else?

lib/server.js
```diff
@@ -234,8 +234,8 @@ class Server {
     }
 
     if (sessionStorageProvider.isAsync()) {
-      sessionStorageProvider.get(sessionId, session =>
-        getSessionCallback(session, null, callback)
+      sessionStorageProvider.get(sessionId,
```
aqrln: Can you please move `sessionId` to a new line?

lib/ws-internal.js
```diff
@@ -149,8 +149,10 @@ module.exports = {
   // see transportCommon.newConnectFn
   //   webSocketConfig - web socket client configuration
   //                     (see connectionFactory)
-  connect: (app, client, webSocketConfig, ...options) =>
-    connect(app, client, webSocketConfig, ...options),
+  connect:
+    (app, client, webSocketConfig, ...options) => connect(
+      app, client, webSocketConfig, ...options
+    ),
```
lundibundi: Maybe just use {} (here and a few other places)? This looks much better imo.
```
connect: (app, client, webSocketConfig, ...options) => { 
  connect(app, client, webSocketConfig, ...options);
}

belochub:
Landed in https://github.com/metarhia/jstp/commit/f530d5c076371b5acbe449103f68022de2a37e23.

PR #355, opened 2018-08-03
build: use UMD for browser bundle

aqrln:
* ~~Publish and export CommonJS modules transpiled with Babel for usage
  with module bundlers~~
* Use UMD for pre-packaged browser bundle *UPD:* and module bundlers
* Update webpack to 4.x

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,14 +25,15 @@
     "jstp-cli": "./bin/cli.js"
   },
   "browser": {
-    "./lib/record-serialization.js": "./lib/record-serialization-fallback.js",
+    "./jstp.js": "./dist/jstp.umd.js",
     "./lib/server.js": false,
     "./lib/net.js": false,
     "./lib/tls.js": false,
     "./lib/socket.js": false,
     "./lib/wss.js": false,
     "./lib/ws.js": "./lib/ws-browser.js",
-    "./lib/simple-auth-policy.js": false
+    "./lib/simple-auth-policy.js": false,
+    "./lib/simple-session-storage-provider.js": false
   },
   "dependencies": {
     "semver": "^5.5.0",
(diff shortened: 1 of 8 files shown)


webpack.config.js
```diff
@@ -7,34 +7,26 @@ const webpack = require('webpack');
...
-    }),
     new webpack.BannerPlugin({
       banner: license,
     }),
   ],
+  node: {
+    net: false,
+    tls: false,
+    // TODO: support WebCrypto API in lib/common.js and uncomment this
+    // crypto: 'empty',
```
aqrln: This will both reduce the browser build size, like, three times, and also improve performance.

package.json
```diff
@@ -25,24 +25,23 @@
...
-    "./lib/ws.js": "./lib/ws-browser.js",
-    "./lib/simple-auth-policy.js": false
+    "./jstp.js": "./dist/cjs/jstp.js",
+    "./dist/cjs/lib/server.js": false,
+    "./dist/cjs/lib/net.js": false,
+    "./dist/cjs/lib/tls.js": false,
+    "./dist/cjs/lib/socket.js": false,
+    "./dist/cjs/lib/wss.js": false,
+    "./dist/cjs/lib/ws.js": "./dist/cjs/lib/ws-browser.js",
+    "./dist/cjs/lib/simple-auth-policy.js": false
```
belochub: I think we can also add: 
```
"./dist/cjs/lib/simple-session-storage-provider.js": false
```
since it can only be used on server side.

belochub:
I've rebased the branch on `master` to resolve the conflicts in `package-lock.json`

belochub:
Landed in https://github.com/metarhia/jstp/commit/0ccd28d764941195336f18e36c9ca75cab902fd4.

PR #356, opened 2018-08-05
test: update Travis config to use the current Node

belochub:
Remove end-of-life Node.js 9 and add the current Node.js 10.


belochub:
Landed in https://github.com/metarhia/jstp/commit/e77080d829fac7d3b61e0f46fa5280008d632c01.

PR #357, opened 2018-08-06
deps: use prettier for code formatting

nechaido:
@belochub @lundibundi @aqrln I would like your opinion on using prettier for automated code formatting in JSTP.


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "requirePragma": false
```
aqrln: It's the default

aqrln:
My approval stands for "+1", but I'd say we must have a univocal approval instead of a simple "no rejections" for this to pass, as this deviates from the house style a bit (when applied organization-wise; for JSTP on its own Prettier actually makes the code style match the other Metarhia projects' one more closely). Also, to avoid inconsistencies throughout the codebase, we should consider making Prettier the house style if @tshemsedinov is fine with it.

nechaido:
@tshemsedinov and @belochub PTAL.

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "proseWrap": "always"
```
belochub: I think we should use the `"preserve"` value for this option (which seems to be the default value) since sometimes wrapping is important and should be preserved (for example take a look at last lines of our `README.md` file).
It will still help with wrapping the long lines in markdown files, but it won't modify other intentionally made wrappings.

belochub:
You should also update `tools/prepare-release.js` file to generate correct markdown files with it.

belochub:
Closing for now due to https://github.com/metarhia/jstp/pull/357#pullrequestreview-144928036.

PR #360, opened 2018-08-07
lib,build: use Web Crypto API in browser

belochub:


lib/common-crypto-fallback.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+/*eslint-env browser*/
```
aqrln: nit: `/* eslint-env browser */`

lib/common-crypto-fallback.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Browser adapter for required functions from node.js crypto module.
```
aqrln: nit: Node.js

lib/common-crypto-fallback.js
```diff
@@ -0,0 +1,42 @@
...
+} else {
+  console.warn(
+    'Web Crypto API is not supported in your browser!',
+    'Using Math.random() instead.'
+  );
+  const getRandomByte = () => Math.floor(0x100 * Math.random());
+  crypto = {
+    randomBytes(count) {
+      const buf = Buffer.alloc(count);
+      for (let i = 0; i < count; i++) {
```
aqrln: Not that essential as it's only three lines of code, but the function below could be reused here.

webpack.config.js
```diff
@@ -35,6 +35,6 @@ module.exports = {
     net: false,
     tls: false,
     // TODO: support WebCrypto API in lib/common.js and uncomment this
```
aqrln: This line is no longer necessary.

lib/common-crypto-fallback.js
```diff
@@ -0,0 +1,30 @@
...
+    window.crypto.getRandomValues(buf);
+  };
+} else {
+  console.warn(
+    'Web Crypto API is not supported in your browser!',
+    'Using Math.random() instead.'
+  );
+  crypto.randomFillSync = (buf) => {
+    for (let i = 0; i < buf.length; i++) {
+      buf[i] = Math.floor(0x100 * Math.random());
```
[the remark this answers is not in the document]
belochub: @lundibundi, I've specifically used `0x100` to emphasize the fact that the resulting number must fit into a byte.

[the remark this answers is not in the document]
aqrln: I don't see any difference one way or the other. `0x100` seems fine as it is, but no strong opinion.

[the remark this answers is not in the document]
lundibundi: Same here, it's just that to me 256 is clearer (the fact that its a byte) than 0x100.

belochub:
Landed in https://github.com/metarhia/jstp/commit/d77434b8cf7b3e94b3a5edb4664d2728c8a41d90.

Issue #361, opened 2018-08-07
test: investigate intermittent resendable-call-from-client failures


aqrln:
Possibly bug or just a flaky test:

```
not ok 30 - test/node/resendable-call-from-client.js # time=241094.551ms
  ---
  timeout: 240000
  file: test/node/resendable-call-from-client.js
  command: /home/travis/.nvm/versions/node/v6.14.3/bin/node
  args:
    - test/node/resendable-call-from-client.js
  stdio:
    - 0
    - pipe
    - 2
  cwd: /home/travis/build/metarhia/jstp
  ...
{
    ok 1 - must connect without errors
    not ok 2 - timeout!
      ---
      signal: SIGTERM
      handles:
        - type: ChildProcess
          events:
            - internalMessage
            - message
        - type: Pipe
        - type: Socket
          events:
            - end
            - finish
            - _socketEnd
            - data
            - error
            - close
      expired: TAP
      stack: |
        emit (node_modules/nyc/node_modules/signal-exit/index.js:77:11)
        process.listener (node_modules/nyc/node_modules/signal-exit/index.js:91:7)
        processEmit (node_modules/nyc/node_modules/signal-exit/index.js:155:32)
        processEmit (node_modules/signal-exit/index.js:155:32)
      test: TAP
      ...
    
    1..2
    # failed 1 of 2 tests
    # time=185627.749ms
    
    not ok 3 - timeout!
      ---
      expired: test/node/resendable-call-from-client.js
      ...
}
```

Evidence: https://travis-ci.org/metarhia/jstp/jobs/413136428

nechaido:
Upon further investigation it look like actual error is:
```
test/node/resendable-call-from-client.js ....events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: channel closed
    at process.target.send (internal/child_process.js:554:16)
    at Object.set (/Users/metarhia/Project/jstp/test/utils/resendable-call/server.js:23:13)
    at Connection.connection.once (/Users/metarhia/Project/jstp/lib/server.js:217:34)
    at Connection.g (events.js:292:16)
    at emitOne (events.js:96:13)
    at Connection.emit (events.js:188:7)
    at Connection._onSocketClose (/Users/metarhia/Project/jstp/lib/connection.js:473:10)
    at emitOne (events.js:96:13)
    at Transport.emit (events.js:188:7)
    at Socket.from.on (/Users/metarhia/Project/jstp/lib/common.js:13:8)
```

[...]

PR #363, opened 2018-08-07
test: fix flaky tests

nechaido:
This should fix the https://github.com/metarhia/jstp/issues/361.

It appears to me that child process was trying to send messages to master process, but at the time, IPC channel was closed. This fix ensures that server is closed after child process and master finished exchanging messages.

### DIFF
--- a/test/node/resendable-call-from-client.js
+++ b/test/node/resendable-call-from-client.js
@@ -44,6 +44,8 @@ function createServer() {
         sessionStorage.set(args[0], args[1]);
         if (initialConnection) {
           call();
+        } else {
+          close();
         }
         break;
       case 'callReceived':
@@ -67,9 +69,10 @@ function call() {
     test.assertNot(error, 'must not encounter an error');
 
     connection.close();
-    server.kill('SIGKILL');
+    close();
     test.end();
   });
+
   server.kill('SIGKILL');
   server = createServer();
 }
@@ -108,4 +111,14 @@ function reconnect(port) {
       }
     });
   });
+}
+
+let calledClose = 0;
+function close() {
+  calledClose++;
+  if (calledClose !== 2) {
+    return;
+  }
+  server.send(['close']);
+  server.disconnect();
 }
--- a/test/node/simple-session-storage-provider.js
+++ b/test/node/simple-session-storage-provider.js
@@ -7,7 +7,7 @@ const SimpleSessionStorageProvider =
 
 const SESSION_LIFETIME = 1000;
 const INTERVAL = 1000;
-const GUARANTEED_DELETION_TIME = INTERVAL * 2;
+const GUARANTEED_DELETION_TIME = INTERVAL * 3;
 const SESSION_ID = 'key';
 const SESSION = {};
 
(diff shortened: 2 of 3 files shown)


test/node/resendable-call-from-client.js
```diff
@@ -109,3 +112,12 @@ function reconnect(port) {
     });
   });
 }
+
+let calledClose = 0;
+function close() {
+  if (++calledClose !== 2) {
```
aqrln: Subjective nit (take it or ignore it): I'm not very fond of such expressions and I'd rather split it into two distinct statements.

test/node/resendable-call-from-client.js
```diff
@@ -44,6 +44,8 @@ function createServer() {
         sessionStorage.set(args[0], args[1]);
         if (initialConnection) {
```
tshemsedinov: Why don't just `initialConnection ? call() : close();` I'd prefer single line for such simple cases or at least
```js
if (initialConnection) call();
else close();
```

tshemsedinov: In addition, we can use `handlers[message]` instead of `switch/case`

test/node/resendable-call-from-client.js
```diff
@@ -109,3 +112,13 @@ function reconnect(port) {
     });
   });
 }
+
+let calledClose = 0;
+function close() {
+  calledClose++;
+  if (calledClose !== 2) {
```
tshemsedinov: why also not a single line?

nechaido: @aqrln asked to split it into two lines see https://github.com/metarhia/jstp/pull/363#discussion_r208611473.

test/node/resendable-call-from-client.js
nechaido: Using `handlers` is a matter for another PR.
IMHO if you use else you must use brackets. As per ternary expression, this PR is already hard to fully understand because of the master-child communication process, so I would like to leave the rest of the code as simple as it can be. If you insist on using the ternary expression, I will rewrite it, but I would prefer not to.

nechaido: Using `handlers` is a matter for another PR. IMHO if you use else you must use brackets. As per ternary expression, this PR is already hard to fully understand because of the master-child communication process, so I would like to leave the rest of the code as simple as it can be. If you insist on using the ternary expression, I will rewrite it, but I would prefer not to.

test/node/resendable-call-from-client.js
aqrln: @nechaido I believe @tshemsedinov is talking about braces, not about the increment

nechaido: In that case, I just don't like `if` statements without braces.

nechaido:
@tshemsedinov we need your answer ASAP as this tests may cause other PR to fail.

test/node/resendable-call-from-client.js
tshemsedinov: Ok

belochub:
Landed in https://github.com/metarhia/jstp/commit/90ebf02068538557098debf1fe13b7969aaeb347.

PR #366, opened 2018-08-09
cli: use new jstp features

nechaido:
Changes:
* rely on a built-in reconnector;
* add possibility to specify verboseness level.


lib/cli/cli.js
```diff
@@ -5,16 +5,49 @@ const EventEmitter = require('events').EventEmitter;
...
+    if (verboseLevel) {
+      const logger = new EventEmitter();
+
+      logger.on('outgoingMessage', (message) => {
+        if (
+          (message.pong || message.ping) &&
+          verboseLevel !== VERBOSE_LEVEL.ALL
+        ) {
+          return;
+        }
```
lundibundi: Could you reverse this and add a logging statement instead of return?

lib/cli/cli.js
```diff
@@ -5,16 +5,49 @@ const EventEmitter = require('events').EventEmitter;
 const CommandProcessor = require('./command-processor');
 const LineProcessor  = require('./line-processor');
 const utils = require('./utils');
+const serde = require('../serde');
+
+const VERBOSE_LEVEL = {
+  NO: 0,
```
lundibundi: Maybe Quiet?

lib/cli/cli.js
```diff
@@ -5,16 +5,49 @@ const EventEmitter = require('events').EventEmitter;
...
+        this.log(`Outgoing message:\n${serde.stringify(message)}`);
+      });
+
+      logger.on('incomingMessage', (message) => {
+        if (
+          (message.pong || message.ping) &&
+          verboseLevel !== VERBOSE_LEVEL.ALL
+        ) {
+          return;
+        }
```
lundibundi: Same as above.

lib/cli/command-processor.js
```diff
@@ -88,31 +88,44 @@ module.exports = class CommandProcessor extends EventEmitter {
...
+    if (!this.cli.connection) {
       callback(new Error('Not connected'));
       return;
     }
+    this.cli.connection.emitRemoteEvent(interfaceName, eventName, args);
+    callback();
+  }
+
+  connect(protocol, host, port, app, interfaces, callback) {
+    this.disconnect(() => {
```
lundibundi: Does this work? (I expect you to get `Not connected` error when you manually connect for the first time)

bin/cli.js
```diff
@@ -3,11 +3,24 @@
...
 const Cli = require('../lib/cli/cli');
 
 let rl = null;
 
+const args = yargs
+  .option('verbose', {
+    alias: 'v',
+    count: true,
+    describe: `Level of verboseness:
+v - print every incoming and outgoing message exept heartbeat messages
```
belochub: s/exept/except

lib/cli/command-processor.js
belochub: @lundibundi, error is ignored in the callback for the `disconnect` method here.

Although, I think it's better to make some kind of internal `_disconnect` method that doesn't check if the connection was established before and call it inside of both `connect` and `disconnect` methods. That way you can also avoid additional `_connect` method creation.

lib/cli/command-processor.js
```diff
@@ -88,31 +88,44 @@ module.exports = class CommandProcessor extends EventEmitter {
...
+  disconnect(callback) {
     if (!this.cli.connection) {
-      callback(new Error('Connection in progress'));
+      callback(new Error('Not connected'));
       return;
     }
-    this.cli.connection.emitRemoteEvent(interfaceName, eventName, args);
+    this.cli.connection.close();
+    this.cli.connection = null;
+    this.cli.connectInitiated = false;
```
belochub: You are not using this field anywhere else.

lib/cli/command-processor.js
nechaido: Yet one question stands: should we close a connection if the initial connection failed or should we let it reconnect?

I say we close it because this is probably because of an error in address, application name or method name.
@belochub @lundibundi what do you think?

lundibundi:
@belochub +1, but I think we can do that in follow-up PR. Also `-r` seems like it actually enables reconect, I'd suggest `-R` for negative aliases or not have this alias at all.

belochub:
@lundibundi, yeah, `-R` is even better.

lib/cli/cli.js
```diff
@@ -5,16 +5,47 @@ const EventEmitter = require('events').EventEmitter;
...
+          (!message.pong && !message.ping) ||
+          verboseLevel === VERBOSE_LEVEL.ALL
+        ) {
+          this.log(`Outgoing message:\n${serde.stringify(message)}`);
+        }
+      });
+
+      this.client = { logger };
+    } else {
+      this.client = null;
```
belochub: You can avoid this branch completely if you set `this.client` to an empty object before this `if`.

bin/cli.js
```diff
@@ -3,11 +3,33 @@
...
+    alias: 'R',
+    type: 'boolean',
+    default: false,
+    describe: 'Disable reconnection',
+  })
+  .option('heartbeat-interval', {
+    alias: 'i',
+    type: 'number',
+    default: 0,
+    describe: 'Heartbeat interval in miliseconds',
```
belochub: s/miliseconds/milliseconds

bin/cli.js
```diff
@@ -2,12 +2,71 @@
...
+  { name: 'heartbeat-interval', alias: 'i', type: Number, default: 0 },
+  { name: 'help', alias: 'h', type: Boolean, default: false },
+];
+
+const usage = [
+  {
+    header: 'JSTP CLI client',
+  },
+  {
+    header: 'Options',
```
belochub: What do you think about changing it to `'Available options'`?

bin/cli.js
```diff
@@ -2,12 +2,71 @@
...
+    optionList: [
```
belochub: You can use the `options` array here if you move `description` and `typeLabel` fields there.

nechaido:
+1 to @lundibundi.

lib/cli/cli.js
```diff
@@ -5,16 +5,52 @@ const EventEmitter = require('events').EventEmitter;
...
+      heartbeatInterval: options.heartbeatInterval,
+    };
+
+    if (options.verbose) {
+      this.client.logger = new EventEmitter();
+
+      this.client.logger.on('outgoingMessage', (message) => {
+        if (
+          (!message.pong && !message.ping) ||
+          options.verbose === VERBOSE_LEVEL.ALL
```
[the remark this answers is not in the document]
nechaido: Fixed.

belochub:
@lundibundi, as an option, we can use `cla-complete` package, which just combines `command-line-args` and `command-line-usage` packages and handles parsing exceptions, but I don't think that it's really required here since such package can be created in minutes and we have only one CLI application to use it with, meaning we can easily write this ourselves and there is no need to abstract it now.
As to the other points of your review, I think these couple of libraries are more flexible in terms of possible configuration and usage in some parts, and, as a result, using them will be more verbose (and requiring some manual checks), than using `yargs`.

lib/cli/cli.js
```diff
@@ -5,16 +5,52 @@ const EventEmitter = require('events').EventEmitter;
...
+          this.log(`Outgoing message:\n${serde.stringify(message)}`);
+        }
+      });
+
+      this.client.logger.on('incomingMessage', (message) => {
+        if (
+          (!message.pong && !message.ping) ||
+          options.verbose === VERBOSE_LEVEL.ALL
+        ) {
+          this.log(`Outgoing message:\n${serde.stringify(message)}`);
```
belochub: Must be "Incoming message" here.

bin/cli.js
```diff
@@ -2,12 +2,83 @@
...
+  if (args.verbose) {
+    args.verbose = args.verbose.length;
+  }
+} catch (error) {
+  console.error(error.message);
+  printHelpAndExit(error.code || 1);
+}
+
+if (args.heartbeatInterval === null) {
+  console.log('You must specify hearbeat interval');
```
belochub: s/hearbeat/heartbeat

belochub: Also, use `console.error` here, please.

lib/cli/cli.js
nechaido: Fixed.

bin/cli.js
```diff
@@ -2,12 +2,83 @@
...
+    alias: 'R',
+    type: Boolean,
+    default: false,
+    description: 'Disable reconnection',
+  },
+  {
+    name: 'heartbeat-interval',
+    alias: 'i',
+    type: Number,
+    default: 0,
```
belochub: The correct name of this property seems to be `defaultValue`.
Although I am not quite sure if we even need to use this property anywhere.

lib/cli/cli.js
```diff
@@ -5,16 +5,52 @@ const EventEmitter = require('events').EventEmitter;
 const CommandProcessor = require('./command-processor');
 const LineProcessor  = require('./line-processor');
 const utils = require('./utils');
+const serde = require('../serde');
+
+const VERBOSE_LEVEL = {
```
lundibundi: maybe `VERBOSENESS_LEVEL` or just `VERBOSENESS`? Though as you want.

nechaido:
@lundibundi, as a matter of fact, it is a PR that must be landed ASAP. so that it shipped with metarhia-jstp@2.

belochub:
@lundibundi, as I said before, we can use `cla-complete` package instead of handling some of the things here, that way the code that you don't want to support will be "handled inside of the cli-args library", it's just going to be another library.

If the aforementioned package isn't good enough in your opinion, I can create another one, and just copy some of the code written by @nechaido in this PR into it, so that it will be handled inside the other package and not this one. That way you don't have to support the code that you don't want to support.

lundibundi:
@belochub and there will be one more package to support, that you will have to update and test somehow. And most likely still not as good looking as `yargs`.
To sum it up, if @nechaido and/or @aqrln agree that it's okay to have this instead of just using `yargs` I'll not object, as this will be better than having our own/your another package to support. (You can probably discuss it with them in person and just post the result/land it if it's okay)

belochub:
@lundibundi, like I said, if you want the package that will handle the things you don't want to handle here, we can either use the `cla-complete` package or create another package with similar functionality.

> there will be one more package to support

I don't understand, can you, please, explain what this other package to support will be?

> you will have to update and test somehow

We don't have any tests for the cli at the moment.

lib/cli/command-processor.js
```diff
@@ -139,14 +135,13 @@ module.exports = class CommandProcessor extends EventEmitter {
         return;
     }
     this.cli.connectInitiated = true;
```
belochub: You should probably remove this line as well.

bin/cli.js
```diff
@@ -2,12 +2,31 @@
...
 
 let rl = null;
 
+const args = yargs
+  .option('verbose', {
+    alias: 'v',
+    count: true,
+    describe: 'Verbosely print incoming and outgoing messages',
+  })
+  .options('no-reconnect', {
```
belochub: It is `options` method here and `option` methods for `'verbose'` and `'heartbeat-interval'`, is there a difference, or is this just a typo?

belochub: After taking a look at `yargs` documentation, it seems that the `options` method has a different contract, requiring to provide the object with options descriptions as a single argument, meaning that the way you use that method here is incorrect.

bin/cli.js
```diff
@@ -2,12 +2,31 @@
...
 const readline = require('readline');
 
 const Cli = require('../lib/cli/cli');
 
 let rl = null;
 
+const args = yargs
+  .option('verbose', {
+    alias: 'v',
+    count: true,
```
belochub: Why do you use `count` option here and not `type` like in other options definitions?

nechaido: I need to use `count` so that when you pass `-vv`, `verbose` will be equal to 2. I do not use `type` here because this option does not have any value associated with but instead, we simply count the number of occurrences of this option.

belochub: I meant that you could've provided `type: 'count'` option instead of using `count` option, to make it consistent with other options definitions.

bin/cli.js
```diff
@@ -2,12 +2,31 @@
...
+    alias: 'R',
+    type: 'boolean',
+    describe: 'Disable reconnection',
+  })
+  .option('heartbeat-interval', {
+    alias: 'i',
+    requiresArg: true,
+    type: 'number',
+    describe: 'Heartbeat interval in milliseconds',
+  }).argv;
```
belochub: In my opinion, we should also use `strict()` here.

PR #368, opened 2018-08-17
test: simplify resendable call tests

nechaido:
This should fix the #361.


aqrln:
I don't frankly think it is a good idea generally, but as I don't see any other solution, I'm not being opposed to it. Perhaps, we should create a separate directory for long-running tests and run only those test cases with `-T`?

nechaido:
@aqrln, unfortunately, this PR doesn't fix the problem.

nechaido:
This PR should be landed after #369 and #370 (to avoid code coverage regression).

lundibundi:
Why would we land it if https://github.com/metarhia/jstp/pull/368#issuecomment-413880631 or is it no longer true? If that's how it is then I'd like to have a separate directory for tests without timeout (I assume configuring per-test timeout is not possible in tab), wdyt?
Also, is 30s really not enough for this test?

aqrln:
@lundibundi it took me a moment to realize what's happening, but the PR title was changed and the branch was replaced with completely different commits. In hindsight, opening a new PR would have been a lot more appropriate to avoid confusion, but at this point it's fine as it is I guess.

nechaido:
@lundibundi I've forgotten to update description in the PR. Increasing timeout doesn't help because the test may not end in some situation.

The current solution is to rewrite the test in a different and much simpler way.

lundibundi:
@aqrln yeah, didn't notice such dramatic change. Edited the comment.

test/node/resendable-call-from-server.js
```diff
@@ -2,51 +2,74 @@
...
 const interfaces = {
   iface: {
-    method: (conn, callback) => {
-      callback(null);
+    method: (connection, callback) => {
+      callsReceivedByClient++;
+      if (callsReceivedByClient === 2) {
+        callback(null);
+      } else if (callsReceivedByClient > 2) {
+        test.fail('Server must not receive more than two call messages');
```
lundibundi: s/Server/Client

test/node/resendable-call-from-server.js
```diff
@@ -2,51 +2,74 @@
...
+    jstp.net.reconnect(connection, port, (error, conn) => {
+      test.error(error, 'must reconnect without errors');
+      connection = conn;
+    });
+  });
+}
+
+test.plan(3);
+
+createServer(() => connect(reconnect));
```
lundibundi: Maybe it's also worth deferring this to nextTick to make sure server had time to send a client a message and the client got it?

test/node/resendable-call-from-client.js
```diff
@@ -11,10 +11,16 @@ const interfaces = {
   iface: {
     method: (connection, callback) => {
       callsReceivedByServer++;
-      if (callsReceivedByServer === 2) {
-        callback(null);
-      } else if (callsReceivedByServer > 2) {
-        test.fail('Server must not receive more than two call messages');
+      switch (callsReceivedByServer) {
```
lundibundi: I think if/else would look better here, though as you want.
Also :+1: for extensive checks.

test/node/resendable-call-from-server.js
```diff
@@ -55,7 +61,7 @@ function connect(callback) {
       test.error(error, 'must connect without errors');
       connection = conn;
       client.session = session;
-      callback();
+      process.nextTick(callback);
```
lundibundi: Why do we have it here? I'd say to put it it the callback itself 
`createServer(() => connect(() => process.nextTick(reconnect)));`

belochub:
Landed in https://github.com/metarhia/jstp/commit/a1a0c0ea70699590429f574dd11040183b6b6e4d.

PR #369, opened 2018-08-17
lib: fix resendable calls being sent twice

belochub:
Due to the callbacks receiving error `ERR_CALLBACK_LOST` being called
synchronously during session restoration, calls with resending were
being resent twice: first time at the moment of such callback call
(leading to saving the call message into the session's buffer) and
second time immediately after the procedure of the session restoration
while resending all of the buffered messages (that by that moment
include unwanted call messages).


belochub:
Landed in https://github.com/metarhia/jstp/commit/2da3c9a4f7743229bc67e68441f6ed1aa4969984.

PR #370, opened 2018-08-17
test: add test for async sessionStorageProvider

nechaido:


test/node/session-provider-async.js
```diff
@@ -0,0 +1,75 @@
...
+    },
+  },
+};
+
+const application = new jstp.Application(APP_NAME, interfaces);
+
+const TIMEOUT = 100;
+const map = new Map();
+const sessionStorageProvider = {
+  set: (sessionId, session) => {
```
lundibundi: Now that I've seen this, I thought about this async set. Ain't there a race condition between `sessionStorageProvider.set` and user reconnect. So if server uses async sessionStorageProvider and user reconnects too fast he may actually not get a session or he will get an outdated (because the session wasn't saved yet) session, wdyt @nechaido @belochub ?
One possible solution may be adding a callback argument to `set` and only `this._clientsBySessionId.delete(connection.session.id);` in the server if we know that it was stored (after callback call).

test/node/session-provider-async.js
```diff
@@ -0,0 +1,75 @@
...
+    second(connection, callback) {
+      callback(null, connection.session.get(TOKEN));
+    },
+  },
+};
+
+const application = new jstp.Application(APP_NAME, interfaces);
+
+const TIMEOUT = 100;
+const map = new Map();
```
lundibundi: Put `map` inside sessionStorageProvider and use function/this? Just a nit.

Issue #377, opened 2018-09-03
build: the UMD bundle is broken ("module.exports is readonly")


aqrln:
Since ea972d63e2858c409b7419bd65508548a6119a08 the UMD bundle fails to load, because `@babel/plugin-transform-runtime` injects ESM `import` statements, hence changing module semantics and making `module.exports` readonly.

PR #378, opened 2018-09-03
build: tell Babel parser the correct source type

aqrln:
Change the default `"sourceType": "module"` to the actual value of
`"sourceType": "script"`. This will prevent Babel from changing
top-level `this` to `undefined` (fwiw), and, more importantly, make
`plugin-transform-runtime` emit `require` statements instead of `import`
statements.

Fixes: https://github.com/metarhia/jstp/issues/377

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,5 +2,6 @@
   "presets": ["@babel/env"],
   "plugins": [
     "@babel/plugin-transform-runtime"
-  ]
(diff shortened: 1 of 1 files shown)


belochub:
Landed in https://github.com/metarhia/jstp/commit/94870c068e620782851a071784853e18717cd275.

PR #381, opened 2018-09-13
ws: use transport similar to the socket ones

belochub:
Previously every JSTP message was sent in a separate WebSocket frame.
This commit changes WebSocket usage so that JSTP message is not limited
to a single WebSocket frame. This is accomplished by making the
WebSocket JSTP transport behave in a way similar to the other available
transports, which is after every message, there is a separator. This
change leads to a WebSocket frame being able to hold multiple JSTP
messages at once or a part of a big JSTP message, that cannot be
transmitted in a single frame. This is a breaking change since it
completely changes the way the messages are transmitted over WebSockets.

### DIFF
--- a/lib/common.js
+++ b/lib/common.js
@@ -143,6 +143,48 @@ class ExpiringMap extends Map {
   }
 }
 
+const utf16singleUnit = 1 << 16;
+
+// https://tools.ietf.org/html/rfc3629#section-3
+const utf8bytesLastCodePoints = {
+  1: 0x007F,
+  2: 0x07FF,
+  3: 0xFFFF,
+};
+
+const utf8split = (str, maxByteCount) => {
+  let utf8length = 0;
+  for (let i = 0; i < str.length;) {
+    const codePoint = str.codePointAt(i);
+
+    // Calculate code point size in bytes when encoded with UTF-8:
+    let utf8codePointSize = 4;
+    for (let byteCount = 1; byteCount <= 3; byteCount++) {
+      if (codePoint <= utf8bytesLastCodePoints[byteCount]) {
+        utf8codePointSize = byteCount;
+        break;
+      }
+    }
+
+    // Calculate code point size in bytes when encoded with UTF-16:
+    const utf16codePointSize = (codePoint >= utf16singleUnit) ? 2 : 1;
+    if (utf8codePointSize > maxByteCount) {
+      return [str.slice(0, i), str.slice(i), utf8length];
+    } else if (utf8codePointSize === maxByteCount) {
+      const splitPoint = i + utf16codePointSize;
+      return [
+        str.slice(0, splitPoint),
+        str.slice(splitPoint),
+        utf8length + utf8codePointSize,
+      ];
+    }
+    maxByteCount -= utf8codePointSize;
+    utf8length += utf8codePointSize;
+    i += utf16codePointSize;
+  }
+  return [str, '', utf8length];
+};
+
 module.exports = {
   forwardEvent,
   forwardMultipleEvents,
@@ -153,4 +195,5 @@ module.exports = {
   rsplit,
   cryptoRandom,
   ExpiringMap,
+  utf8split,
 };
--- a/lib/common.js
+++ b/lib/common.js
@@ -143,7 +143,7 @@ class ExpiringMap extends Map {
   }
 }
 
-const utf16singleUnit = 1 << 16;
+const utf16singleUnit = (1 << 16) - 1;
 
 // https://tools.ietf.org/html/rfc3629#section-3
 const utf8bytesLastCodePoints = {
@@ -167,7 +167,7 @@ const utf8split = (str, maxByteCount) => {
     }
 
     // Calculate code point size in bytes when encoded with UTF-16:
-    const utf16codePointSize = (codePoint >= utf16singleUnit) ? 2 : 1;
+    const utf16codePointSize = (codePoint > utf16singleUnit) ? 2 : 1;
     if (utf8codePointSize > maxByteCount) {
       return [str.slice(0, i), str.slice(i), utf8length];
     } else if (utf8codePointSize === maxByteCount) {
--- a/lib/internal-constants.js
+++ b/lib/internal-constants.js
@@ -2,4 +2,6 @@
 
 module.exports = {
   WEBSOCKET_PROTOCOL_NAME: 'metarhia-jstp',
+  MAX_MESSAGE_SIZE: 8 * 1024 * 1024,
+  SEPARATOR: '\0',
 };
--- a/lib/ws-browser.js
+++ b/lib/ws-browser.js
@@ -5,8 +5,12 @@ const EventEmitter = require('events').EventEmitter;
 
 const mdsf = require('mdsf');
 
+const { utf8split } = require('./common');
 const constants = require('./internal-constants');
 const transportCommon = require('./transport-common');
+
+const SEND_INTERVAL = 50;
+const BUFFER_SIZE_LIMIT = 64000;
 
 // W3C WebSocket transport for JSTP.
 //   socket - WebSocket instance
@@ -17,16 +21,26 @@ class Transport extends EventEmitter {
     super();
 
     this.socket = socket;
+    this._receiveBuffer = '';
+    this._sendBuffer = [];
+    this._closed = false;
 
     this.socket.onmessage = message => {
       this._onMessage(message);
     };
 
-
     ['close', 'error'].forEach(event => {
       this.socket.addEventListener(event, (...args) => {
         this.emit(event, ...args);
       });
+    });
+
+    this._sendInterval = setInterval(() => {
+      this._sendBufferedMessages();
+    }, SEND_INTERVAL);
+
+    this.once('close', () => {
+      clearInterval(this._sendInterval);
     });
   }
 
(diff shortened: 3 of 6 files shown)


lib/ws-browser.js
```diff
@@ -36,6 +49,29 @@ class Transport extends EventEmitter {
...
+    let toSend = '';
+    while (this._sendBuffer.length !== 0) {
+      const message = this._sendBuffer[0];
+      if (message.length + bufferedAmount <= BUFFER_SIZE_LIMIT) {
+        toSend += message;
+        bufferedAmount += message.length;
+        this._sendBuffer.shift();
+      } else {
+        const chunkLength = BUFFER_SIZE_LIMIT - bufferedAmount;
+        if (chunkLength !== 0) {
```
lundibundi: I understand that this algorithm is made so that chunkLength is always >= 0 though I'd still prefer a >0 check here.

belochub: If the `bufferedAmount` value is bigger than `BUFFER_SIZE_LIMIT` value, it must be considered an error, not a normal situation. Considering this, I think it's better to add the check for `bufferedAmount > BUFFER_SIZE_LIMIT` (or `chunkLenght < 0`) that will throw in such a situation.

lib/internal-constants.js
```diff
@@ -2,4 +2,6 @@
 
 module.exports = {
   WEBSOCKET_PROTOCOL_NAME: 'metarhia-jstp',
+  MAX_MESSAGE_SIZE: 8 * 1024 * 1024,
```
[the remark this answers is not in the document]
belochub: We can do it, but it has nothing to do with this PR.

belochub:
@lundibundi the solution you proposed wouldn't allow to send multiple JSTP messages in one WebSocket frame except for some really rare cases where smaller JSTP messages (that can fit into one WS frame) are buffered instead of being sent immediately due to a bigger JSTP message (that can't fit into one WS frame) being transmitted at that time.

lib/ws-browser.js
lundibundi: Well, I don't see how's that situation possible so I'm not sure if we should check for that.
All I wanted is for this check to be a bit more clear when you read the code, meaning - if chunkLen is bigger than 0 then we have something to write.

belochub: This situation is possible if someone using the package will access the WebSocket connection directly and just call the `send()` method on it.

lundibundi:
Why is it important to send them in one frame then? I thought the idea was to avoid flooding the WS with messages so it won't drop the connection upon sending lots of messages.

Basically I don't like the idea of running 50ms interval for each WS connection the server has, i.e. if we have only 1k connections that don't send anything that would still result in 1k callbacks being evaluated every 50ms and is basically wasting resources. Therefore I'd suggest to either:
* use the scheme I suggested and only run interval when needed
* use your scheme but run the interval only when there is something to send and remove it otherwise, this will at least somehow offload the unneded work

Maybe @nechaido or @aqrln has any other ideas?

belochub:
>I thought the idea was to avoid flooding the WS with messages so it won't drop the connection upon sending lots of messages.

WebSocket does not drop the connection when you are sending lots of messages, and this PR was not created to avoid flooding it with messages.
The problem that exists is that, according to the W3C standard for WebSockets, `send(data)` method operates in such way:
>if the data cannot be sent, e.g. because it would need to be buffered but the buffer is full, the user agent must flag the WebSocket as full and then close the WebSocket connection

This makes it impossible to send JSTP message bigger than browser's buffer for WebSocket frames (which is usually 64KiB) since every JSTP message is sent in a single frame. This PR solves this problem by sending bigger messages in multiple WebSocket frames sequentially.

The way of handling the `bufferedAmount` is based on this basic example from the W3C standard for WebSockets: https://html.spec.whatwg.org/multipage/web-sockets.html#dom-websocket-bufferedamount

As a positive side effect, this way makes it possible to transmit multiple JSTP messages in a single WebSocket frame.

belochub:
@lundibundi 
Also, about
>running 50ms interval for each WS connection the server has

Take a closer look at the code, the `setInterval` function is only used in browser, not on the server.

belochub:
Unfortunately, the implementation had one important problem before, that of `bufferedAmount` attribute returning the number of bytes of the UTF-8 text being sent, not UTF-16 that is used in strings in JavaScript, so I've updated the implementation to account for that.

lib/ws-browser.js
```diff
@@ -36,15 +94,51 @@ class Transport extends EventEmitter {
...
+        toSend += message;
+        bufferedAmount += utf8messageLength;
+        this._sendBuffer.shift();
+      } else {
+        const chunkLength = BUFFER_SIZE_LIMIT - bufferedAmount;
+        if (chunkLength < 0) {
+          this.socket.close();
+          throw new Error('WebSocket buffer is overflown');
+        }
+        if (chunkLength !== 0) {
```
lundibundi: As I mentioned before I'd prefer for this to be `> 0` check.

belochub: There is no need to make a `> 0` check since I check if the value is `< 0` and then exit before this line in that case, and because of that, checking for `!== 0` here covers all of the other values which are `> 0`.

belochub:
@lundibundi 
I think it can be simplified even more if we use a string buffer instead of an array of strings, what do you think about that?

lundibundi:
I'm not sure about the performance implications of that, though if you think it'll have better performance than an Array feel free to implement it that way.

:thinking: What do you think of storing strings of maxLength right in the `send` method: upon `send` you check the length of the last element in the array and if it's less than maxLength you append to it the remaining amount and then push everything else as the next element of an array. This may lessen the burden of 'interval' of going through elements and appending/slicing them manually in one place instead making it as user `send`s messages, wdyt?

belochub:
@lundibundi I don't think we can land it now, since this is missing some tests, and we should also wait for a review from @aqrln.

lib/ws-browser.js
```diff
@@ -8,6 +8,54 @@ const mdsf = require('mdsf');
 const constants = require('./internal-constants');
 const transportCommon = require('./transport-common');
 
+const SEND_INTERVAL = 50;
+const BUFFER_SIZE_LIMIT = 64000;
+
+const utf16singleUnit = 1 << 16;
```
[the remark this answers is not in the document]
belochub: I don't think it will simplify anything, we will still have to convert the strings that we receive from `mdsf.stringify()` to binary UTF-8 representation.

test/node/common-utf8split.js
```diff
@@ -0,0 +1,204 @@
...
+        result: ['💓💕💖💗💝', '', 20],
+      },
+    ];
+    runTestCases(test, string, testCases);
+    test.end();
+  }
+);
+
+test(
+  'must correctly split strings with code points using different counts of ' +
```
nechaido: Maybe:
```js
  'must correctly split strings with code points using' +
    'variable length UTF-8 code units'
```

lib/ws-browser.js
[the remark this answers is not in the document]
aqrln: Yeah but we have Buffer polyfill anyway, so it's a one-liner.

test/node/common-utf8split.js
belochub: That will be incorrect since UTF-8 code units can't have variable length, they are always 8 bit long.

lib/ws-browser.js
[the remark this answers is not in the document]
belochub: Using binary frames will also disallow inspecting the WebSocket frames in browser's dev tools and won't make it easy to split UTF-8 messages if we want to avoid the possibility of breaking a single Unicode character between frames.
I am not sure if these are important to consider, though.

lib/common.js
```diff
@@ -145,20 +145,28 @@ class ExpiringMap extends Map {
...
-    } else if (codePoint < 0x800) {
-      utf8codePointSize = 2;
-    } else if (codePoint < 0x10000) {
-      utf8codePointSize = 3;
-    } else {
-      utf8codePointSize = 4;
+
+    // Calculate code point size in bytes when encoded with UTF-8:
+    let utf8codePointSize = 4;
+    for (let byteCount = 1; byteCount <= 3; byteCount++) {
```
lundibundi: Though no need to change it as the way it is done below is much more cryptic 🙃.
```js
let byteCount = 0;
while (++byteCount < 4 && codePoint > utf8bytesLastCodePoints[byteCount]);
const utf8codePointSize = byteCount;
```
Good ol' C =) We can even include `4` in the utf8bytesLastCodePoints and go even further.

belochub: What's the point of including `4` in the `utf8bytesLastCodePoints`?
You can't get the `codePoint` higher than `0x10FFFF` which is the maximal Unicode code point that can be represented by 4 UTF-8 code units.

lundibundi: Of I forgot it's `undefined` check. Then just
```js
let byteCount = 0;
while (codePoint > utf8bytesLastCodePoints[++byteCount]);
const utf8codePointSize = byteCount;
```
:upside_down_face:

lib/ws-browser.js
```diff
@@ -36,15 +50,53 @@ class Transport extends EventEmitter {
...
 
+  _sendBufferedMessages() {
+    let bufferedAmount = this.socket.bufferedAmount;
+    let toSend = '';
+    while (this._sendBuffer.length !== 0) {
+      const message = this._sendBuffer[0];
+      const maxChunkLength = BUFFER_SIZE_LIMIT - bufferedAmount;
+      if (maxChunkLength < 0) {
+        this.socket.close();
+        throw new Error('WebSocket buffer is overflown');
```
aqrln: But if `this.socket.bufferedAmount > BUFFER_SIZE_LIMIT` and there's no exception, doesn't it mean that the actual limit in a particular browser is higher than the value of that constant and we're just dropping the connection for no reason? Instead, we can increase `BUFFER_SIZE_LIMIT` to the value of `this.socket.bufferedAmount` and try again at the next timer tick.

aqrln: I mean, that's the only way for the condition to hit.

belochub: @aqrln when using only function `_sendBufferedMessages` to send anything over the WebSocket it is not possible to reach the state where this condition will be met. The only way to get this error is to use the WebSocket socket directly and call the `send()` method by accessing the `socket` field of the instance of this W3C WebSocket `Transport` class.
The reasoning behind this condition was explained before in [this thread](https://github.com/metarhia/jstp/pull/381#discussion_r217608269).

belochub:
Closing in favor of https://github.com/metarhia/jstp/pull/383.

PR #383, opened 2018-10-01
ws,server: change max frame and message size

belochub:
Change the maximally allowed frame and message size for WS server to
a maximal size of JSTP message allowed (8 MiB).

This fixes the same issue as https://github.com/metarhia/jstp/pull/381 with the least amount of changes, thus I'd prefer to land this PR instead of https://github.com/metarhia/jstp/pull/381.

### DIFF
--- a/lib/internal-constants.js
+++ b/lib/internal-constants.js
@@ -2,4 +2,5 @@
 
 module.exports = {
   WEBSOCKET_PROTOCOL_NAME: 'metarhia-jstp',
+  MAX_MESSAGE_SIZE: 8 * 1024 * 1024,
 };
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -6,9 +6,9 @@ const mdsf = require('mdsf');
 
 const transportCommon = require('./transport-common');
 const common = require('./common');
+const { MAX_MESSAGE_SIZE } = require('./internal-constants');
 
 const SEPARATOR = Buffer.alloc(1);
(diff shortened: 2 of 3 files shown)


lundibundi:
Wouldn't it be possible to actually send messages (from client) of size more than 8MB with #381 (once we make MAX_MESSAGE_SIZE on the server configurable)? While this change would still require a user to manually split their messages.
IMO we should land both of those PRs IIUC.

belochub:
@lundibundi 
It is possible to send messages of size more than 8MB after this PR lands because the limit is set to 8MiB. But regardless of whether we land the https://github.com/metarhia/jstp/pull/381 or not, it will be possible for the client to send messages bigger than 8MiB if we make the `MAX_MESSAGE_SIZE` configurable, there still will be no need to split the messages on the client side.

aqrln:
@lundibundi there's some context missing in #381 because at some point I stopped reviewing it on GitHub and brought it up in person with @belochub. Browsers don't limit message size to 64 KiB or whatever, the purpose of `bufferedAmount` it to allow you to implement rate limiting (and if you check its value for anything else than to compare it with `0`, then how the value changes over time is what is important, not a momentary value), and the only reason connections were dropped is because the server dropped them due to its default settings.

lundibundi:
@aqrln thanks for the explanation. LGTM then.

belochub:
Landed in https://github.com/metarhia/jstp/commit/c008830ffdff7d96fbae3f9ff513de3960d42aa7.

PR #386, opened 2018-11-27
conn: make reconnection consistent

belochub:
Completely disable reconnection before establishing a JSTP connection
successfully for at least one time.

Before, reconnection retries were implicitly happening before
successfully establishing JSTP connection in a case when transport
(tcp/tls/ws) connection was successfully established, but JSTP handshake
failed. At the same time, reconnection was disabled whenever it was not
possible to establish a transport connection to the server.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -56,6 +56,7 @@ class Connection extends EventEmitter {
 
     this._heartbeatTimer = null;
 
+    this.hasConnected = false;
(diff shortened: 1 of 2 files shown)


belochub:
Landed in https://github.com/metarhia/jstp/commit/dd0d80a7641b904e830a316eab81d6d86158b12d.

PR #387, opened 2018-11-27
conn: return error when handshake is not received

belochub:
Return error from `Connection#handshake()` method when connection is
closed before handshake response is received. Also, make this method's
`callback` argument mandatory.

This depends on https://github.com/metarhia/jstp/pull/386 and https://github.com/metarhia/jstp/pull/389

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -220,12 +220,7 @@ class Connection extends EventEmitter {
     }
 
     if (version && !semver.validRange(version)) {
-      const error = new Error('Invalid semver version range');
-      if (callback) {
-        callback(error);
-      } else {
-        this.emit('error', error);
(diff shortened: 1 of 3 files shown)


lib/connection.js
```diff
@@ -56,6 +56,7 @@ class Connection extends EventEmitter {
 
     this._heartbeatTimer = null;
 
+    this.connectedAtLeastOnce = false;
```
lundibundi: Also, possible names: `hadConnected`, `connectedOnce`.

belochub:
Landed in https://github.com/metarhia/jstp/commit/4caa511751e94fe0bf9dcc272a3dd4fdfac6d198.

PR #388, opened 2018-11-27
test: fix invalid test

nechaido:
It appears that port was actually undefined, which is not the desired behavior in a given tests.


belochub:
Landed in https://github.com/metarhia/jstp/commit/306fa65e0b4c0954a297932a81d64af8fa6d4208.

PR #389, opened 2018-11-27
 lib: fix unhandled error events on connection

belochub:
Fix error events on `Connection` object being unhandled inside the
`Transport#connect()` method, thus leading to the errors being thrown
instead of reported to the callback.

This depends on https://github.com/metarhia/jstp/pull/386


belochub:
Landed in https://github.com/metarhia/jstp/commit/85c18c61f8231116acd212e3896eaed6542f5a54.

PR #390, opened 2018-11-30
cli: add persistent history implementation

belochub:


bin/cli.js
```diff
@@ -60,4 +74,10 @@ rl.on('close', () => finish());
 
 rl.on('SIGINT', () => finish());
 
-rl.prompt(true);
+setupHistory(rl, (err, message) => {
+  if (err) {
+    if (message) log(message);
+    if (args.verbose) log(err.stack);
```
lundibundi: I'd prefer to use `VERBOSENESS_LEVEL` constants or at least make this explicit i.e. `args.verbose !== 0`.

lib/cli/history.js
```diff
@@ -0,0 +1,99 @@
...
+  createFile();
+
+  function createFile() {
+    fs.open(resolvedPath, 'a+', 0o0600, (err, fd) => {
+      if (err) {
+        callback(err,
+          'Error: could not open history file\nHistory will not be persisted');
+        return;
+      }
+      fs.close(fd, onClose);
```
lundibundi: Why close and then reopen in `loadHistory` if we could just read from this `fd`?

belochub:
Landed in https://github.com/metarhia/jstp/commit/da5bc6a8553904d0a01bef2444cc037e980ff128.

PR #392, opened 2018-12-07
ws,server: avoid forwarding 'clientError' event

belochub:
Avoid forwarding 'clientError' event on the WebSocket server as an
'error' event. That behavior was causing server crashes and could
possibly lead to hanging HTTP connections.


belochub:
@aqrln, probably yes, I am not sure if that's the only way to get this event, though I still don't think that it is entirely correct to forward server's `'clientError'` event as an `'error'` event since it possibly leads to server crashes.

belochub:
Landed in https://github.com/metarhia/jstp/commit/f2a756403a7ec9dea0ddf9672cc08b3a788b7a71.

PR #393, opened 2018-12-12
lib: change the default reconnector behavior

belochub:
Avoid trying to reconnect for additional one time after the connection is intentionally closed in some cases.


belochub:
Landed in https://github.com/metarhia/jstp/commit/316e75fb5d530f86c745abb8430cab4806d20234.

PR #404, opened 2019-01-14
lib: remove unsupported Web Crypto API console warning

aqrln:
Currently JSTP prints a warning about using Math.random() to browser
console unless the browser supports Web Crypto API.  However, the
purpose of this warning isn't clear since it is not actionable and
provides zero value for end users.  Although an extra warning message
that users won't probably see anyway isn't a big issue, it becomes
unnecessary noisy in fake browser environment used for tests (in
particular, I get two warnings for each Jest test case, i.e., about two
thirds of all output).


belochub:
Landed in https://github.com/metarhia/jstp/commit/a435227c2db86129213f762966dcf92666f8c09d.

PR #411, opened 2019-02-05
lib: fix invalid args handling in reconnectFn()

belochub:
Before, when the transport was passed to the `reconnectFn()`, it was
then not removed from the argument list before passing it on to the
`transport.reconnect()` method.

Co-authored-by: Dmytro Nechai <nechaido@gmail.com>

### DIFF
--- a/test/node/regress-gh-411.js
+++ b/test/node/regress-gh-411.js
@@ -1,0 +1,43 @@
+'use strict';
+
+const test = require('tap');
+
+const jstp = require('../..');
+
+const APP_NAME = 'APP_NAME';
+
+const reconnector = (connection, reconnectFn) => {
+  if (!connection.closedIntentionally) {
+    // Explicit transport passing is the main purpose of this test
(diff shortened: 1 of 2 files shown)


test/node/regress-gh-411.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+};
+
+test.beforeEach(done => {
+  server = jstp.net.createServer(serverConfig);
+  server.listen(0, () => {
+    const port = server.address().port;
+    jstp.net.connect(
+      APP_NAME,
+      { reconnector },
```
lundibundi: I'd prefer to have an explicit mention that this is what is actually the core of the test. Because based on `test.test` it looks like we are testing that reconnect is called after the transport destroy. Maybe it's also worth adding a short description at the top.

belochub:
Landed in https://github.com/metarhia/jstp/commit/3f5d89066956ab56360c4ff173c096d763cf375a.

Issue #413, opened 2019-02-06
ExpiringMap bug


ykolomiets:
Hi, guys!
I think you have a bug here.
After removing an element from a map, you don't remove the same key from _deletionTimes,
so if you set new element with the same key it will be removed on next _purge. If set only unique keys this is not critical, but _deletionTimes just grows

https://github.com/metarhia/jstp/blob/3f5d89066956ab56360c4ff173c096d763cf375a/lib/common.js#L126-L133

belochub:
Hi, thanks for reporting this!
I see where the problem is and what the possible negative effects are, fix is on the way.

PR #414, opened 2019-02-06
lib: fix ExpiringMap internal storage growing

belochub:
Internal storage of ExpiringMap used to store key deletion times was not
cleared after deleting corresponding keys, leading to its indefinite
growth.

Fixes: https://github.com/metarhia/jstp/issues/413


belochub:
Landed in https://github.com/metarhia/jstp/commit/700776e248e40bb0750d1a3105263be6a509140f.

PR #415, opened 2019-02-07
cli: improve args serialization

belochub:
Drop the brackets around the callback/event arguments and output
`'<empty>'` when there are no arguments.

### DIFF
--- a/lib/cli/utils.js
+++ b/lib/cli/utils.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const mdsf = require('mdsf');
+
 const utils = {};
 module.exports = utils;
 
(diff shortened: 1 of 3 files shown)


lib/cli/utils.js
```diff
@@ -93,3 +95,15 @@ utils.iterativeCompletion = (inputs, depth, completer) => {
...
+  let res = '\n';
+  if (args.length === 0) {
+    res = '<empty>';
+  } else if (typeof indent === 'string' || Number.isInteger(indent)) {
+    res += args.map(v => mdsf.stringify(v, null, indent)).join(',\n');
+  } else {
+    res += mdsf.stringify(args, null, indent).slice(1, -1);
+  }
+  return res;
+};
```
tshemsedinov:
```js
utils.serializeArgs = (args, indent) => {
  if (args.length === 0) return '<empty>';
  if (typeof indent === 'string' || Number.isInteger(indent)) {
    return '\n' + args.map(v => mdsf.stringify(v, null, indent)).join(',\n');
  }
  return '\n' + mdsf.stringify(args, null, indent).slice(1, -1);
};
```

belochub:
Landed in https://github.com/metarhia/jstp/commit/dc698b3250c99679bef1414a23991d667cebfe2e.

PR #425, opened 2019-04-09
lib: allow passing additional info to RemoteError

belochub:


nechaido:
IMHO, `RemoteError.toJSTP` should also return `info` if it is present.

lib/errors.js
```diff
@@ -91,7 +96,11 @@ class RemoteError extends Error {
   //   array - array in the form of [code, description]
   //
   static fromJstpArray(array) {
-    return new RemoteError(array[0], array[1]);
+    return new RemoteError(
+      array[0],
+      array[1],
+      array.length > 2 ? array.slice(2) : null
```
lundibundi: Perhaps we can just pass an array as a third argument to avoid destructuring in `toJSTP` (and simplify this to `array[2]`)?

belochub: I don't think that it will be better from the API user's point of view. It'll become possible to transmit an empty array having no information in it in place of the `info` over the network. And having the nested array in the message makes it somewhat harder to read for a user in my opinion.
IIUC, you don't insist on this change since you've approved the PR, so I am going to proceed with landing this PR.

belochub:
Landed in https://github.com/metarhia/jstp/commit/d23e4855c744ff1c7e940cf2608fabc3adfda822.

PR #427, opened 2019-05-15
Drop support for Node.js versions older than 8.0

belochub:
Also, add Node.js 12 to the Travis CI configuration, update test for `async` application methods, and apply some syntactic sugar that was unavailable before.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,6 +3,8 @@ image: Visual Studio 2015
 environment:
   matrix:
     - nodejs_version: 8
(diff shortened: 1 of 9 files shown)


nechaido (approved):

SemenchenkoVitaliy (approved):

lundibundi (approved):

tshemsedinov (approved):

tshemsedinov:
Why not landing? I think we can even drop all version `>= 10.16.0`

PR #433, opened 2019-12-07
Update dependencies (fix security)

tshemsedinov:


tshemsedinov:
I think we can land this after https://github.com/metarhia/jstp/pull/427
@belochub

